PR #1, opened 2015-11-16
Initial implementation

masylum:
#### :tophat: What? Why?

First prototype to support https://github.com/ipfs/specs/tree/master/repo
Implemented part of `fs-repo` adapter by using `fs-blob-store`, paving the way
for browser-compatible blob stores (@somebody32 working on it)

#### :ghost: GIF

^ streams


masylum:
@somebody32 I removed babel because it was slow and got tired of it, the code looks all-right anyway

masylum:
@diasdavid would be nice if you could take a look at the initial proposal to be sure I'm on the right way

src/adaptors/fs.js
```diff
@@ -0,0 +1,67 @@
...
+  this.LOCK_PATH = path + '/repo.lock'
+}
+
+BlobStore.prototype = {
+  read: function (key, cb) {
+    var store = this.store
+    var LOCK_PATH = this.LOCK_PATH
+
+    function onFinish (buff) {
+      lockFile.unlock(LOCK_PATH, function () {
```
masylum: missing error handling

src/adaptors/fs.js
```diff
@@ -0,0 +1,88 @@
...
+    rs.pipe(concat(function (buff) {
+      cb(null, buff.toString('utf8'))
+    }))
+  },
+
+  append: function (key, content, cb) {
+    return this._write(key, content, 'a', cb)
+  },
+
+  write: function (key, content, cb) {
```
diasdavid: It would be nice to also offer a stream interface

README.md
```diff
@@ -1,2 +1,103 @@
...
+repo.repo().read(function (err, content) {
+  console.log(err, content)
+})
+
+repo.repo().write('foo', function (err) {
+  console.log(err)
+})
+```
+
+## Adaptors
```
diasdavid: Maybe keep calling them blob-stores and point to abstract-blob-store?

diasdavid:
Seems you are in the right direction :) I like that the interface revolves the several types of data stored in the Repo. It would be good to start having tests for the experiments you are doing, so other people feel confident to PR too.

You had some thoughts on how repo should be structured on disk, might wanna share with @whyrusleeping before 0.4.0 release, since 0.4.0 has a different repo structure already, it will be the perfect time for other significant migrations.

README.md
masylum: the issue I'm having is that those adaptors can't just be blob-stores. There is some extra plumbing that needs to be done around. For instance the `fs-repo` need locking through a `repo.lock` file which may be implemented completely different on a `IndexedDB` or `localstorage` implementation. Also, I may want to list all the blocks in a repository, that's not something doable within a blob-store and will have different implementations.

src/adaptors/fs.js
masylum: good point, initially I was returning the writeable stream but it got lost when implementing the locks. Will create a richer API as I move forward.

masylum:
Is there anywhere I can see what's planned in the 0.4.0 release? I've reached a point where I will have to start reverse-engineering the go implementation though since specs and reality don't seem to match (opened https://github.com/ipfs/specs/issues/41)

jbenet:
@masylum I'm sorry about that, I really need to fix the spec there to
reflect go. :c

I think we can move the nesting to be 3 levels in go-ipfs but some other
stuff is better than I need to materialize here.

If anybody wants to help me writing this stuff up it would be sweet.

whyrusleeping:
tv left some notes here that may help: https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/fsrepo.go#L356

masylum:
I moved forward with the prefix as it is in go-ipfs. If the spec materializes in some other way is easy for me to change it back.

Added a `memory-repo` dummy implementation which will allow me to start testing easily. It will also help me define a `fs` agnostic API (I still have my doubts whether `repo.lock` should belong to repo or fs-repo).

I also will need some guidance with the rest since I'm not sure I understand the `datastore` folder.

jbenet:
@masylum sounds good to me. 

re the repo.lock, i responded in the spec. gist is that yes, that probably should be removed from `repo` spec.

diasdavid:
@masylum you have write perms on this repo, mind if you do a PR to a branch in this repo and then a PR to master, so we can work together without having to PR the PR the PR :)

diasdavid:
I went ahead and moved the PR to be a branch on the main repo, so it enables easier collaboration - https://github.com/ipfs/js-ipfs-repo/pull/2. Closing this one.

PR #2, opened 2015-12-10
Bring @masylum PR and @diasdavid PR to be a PR from a ipfs/js-ipfs-repo branch to master

diasdavid:
Makes contribution easier :)


README.md
```diff
@@ -1,2 +1,113 @@
...
+Options:
+
+  - `adaptor`: String with the adaptor. Defaults to `fs`
+
+### `#version`
+
+Read/Write the version number of that repository.
+
+```js
+repo.version().read(function (err, num) {
```
diasdavid: any specific reason why `version()` has to be a func?

masylum: it was easier to develop but we can indeed have a better API

src/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+    if (!adaptor) {
+      throw new Error('Adaptor "' + this.options.adaptor + '" not supported')
+    }
+
+    return adaptor
+  },
+
+  api: function () {
+    return stores.config(this.store)
```
masylum: typo here: `stores.api` (my fault)

README.md
masylum: I'm "injecting" the store on each of those (https://github.com/ipfs/js-ipfs-repo/pull/2/files#diff-1fdf421c05c1140f6d71444ea2b27638R36). I can improve that I guess.

src/adaptors/memory-repo/index.js
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Read a blob on a given path
+   *
+   * @param {String} key
+   * @param {Function} cb
+   * @return {ReadableStream}
+   */
+  readWithoutLock: function (key, cb) {
+    return this.read(key, cb)
+  },
```
masylum: This is something that I'm not fully sure about. Should be locks on the API?

src/stores/index.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  version: require('./version'),
+  config: require('./config'),
+  // logs: require('./logs'),
```
masylum: those are meant to disappear from the repos

README.md
diasdavid: I've added a `load` method, so that we can have a pre state where we can check if a repo exists, but without loading some of the variables to memory (things like version or config that can be read once), and also start the remaining stores, avoiding the need to have inject the 'store' for each call :)

tests/repo-test.js
```diff
@@ -0,0 +1,69 @@
...
+        done()
+      })
+    })
+
+    it('set version', function (done) {
+      done()
+    })
+  })
+  describe('blocks', function () {})
+})
```
masylum: wouldn't be easy to test with the memory implementation?

src/index.js
```diff
@@ -0,0 +1,49 @@
...
+      return false
+    }
+  },
+
+  load: function () {
+    if (this.loaded) { return }
+    var Adaptor = this._chooseAdaptor()
+    this.store = new Adaptor(this.root_path)
+
+    this.api = stores.config(this.store)
```
masylum: `stores.api`

README.md
masylum: better. Any reason why is this not in the constructor?

src/adaptors/memory-repo/index.js
diasdavid: I'm not sure if we will want to use it directly, but won't hurt having them available/thought through :)

diasdavid:
I'm going to merge this and publish to npm, so that we can start jsipfs with a good precedent of having tests passing on travis, as we ship items from the Roadmap https://github.com/ipfs/pm/pull/71

Issue #9, opened 2015-12-10
IPFS Repo Feature - Implement `locks`


diasdavid:
> IPFS implementations may use multiple processes, or may disallow multiple processes from using the same repo simultaneously. Others may disallow using the same repo but may allow sharing datastores simultaneously. This synchronization is accomplished via locks.

Full information: https://github.com/ipfs/specs/blob/fix-repo/repo/README.md#locks
Complete Repo Spec - https://github.com/ipfs/specs/blob/fix-repo/repo/README.md

diasdavid:
I'm thinking that it would be nice to have a `store`, that builds on top of `abstract-blob-store`, introduces features as locks, but doesn't necessarily require for every blob-store implementation to be [crafted by hand](https://github.com/ipfs/js-ipfs-repo/tree/master/src/adaptors).

This will require things like upgrading `lockfile` to support abstract-blob-store

diasdavid:
@whyrusleeping just to make sure I get this right, it is supposed for ipfs-repo to support more than one type of store at the same time? Looking at a regular ipfs repo, seems like datastore is using levelDB and the rest is just straight fs. (What I'm trying to understand is if I can have a clean interface for this things so that it doesn't get super complicated to make it work in the browser as well)

diasdavid:
Ok, going to change how ipfs-repo is changed a bit, so that we support multiple types of storage mechanism depending on the type of data being stored. 

I'm going to follow the diagram:
![fullsizerender](https://cloud.githubusercontent.com/assets/1211152/11735896/c7e19b90-9f7b-11e5-9262-b7bc81799be8.jpg)

With this, we can select the best datastore for each type of data, while still keeping the option of using levelDB/IndexedDB without any big changes, so it works nicely in the browser

- https://github.com/maxogden/abstract-blob-store
- https://github.com/mafintosh/fs-blob-store
- https://github.com/substack/idb-blob-store

jbenet:
Some clarity on the question about stores:

go-ipfs history:
- we started with leveldb
- we moved blocks to flat-fs
- we are in process of moving things off leveldb entirely (like pinset in dev0.4.0)
- my goal is to get rid of leveldb entirely

regardless, it is the case that there is  "one datastore", it's just split in two underneath the hood:
- https://github.com/ipfs/go-ipfs/blob/master/repo/fsrepo/fsrepo.go#L97
[...]

PR #11, opened 2015-12-11
WIP - locks

diasdavid:
-> https://github.com/ipfs/js-ipfs-repo/issues/9

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,6 @@
 var stores = require('./stores')
 var extend = require('xtend')
 var fs = require('fs-blob-store')
-var path = require('path')
-var ncp = require('ncp')
 
 exports = module.exports = Repo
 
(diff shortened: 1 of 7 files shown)


diasdavid:
Travis, you are drunk, go home. There is no .read on line 57 and you are looking at the wrong commit (03afbce !== 8864197)

src/index.js
```diff
@@ -1,68 +1,72 @@
...
-      return !!fs.statSync(this.root_path)
+      return !!fs.statSync(this.repoPath)
     } catch (err) {
       return false
     }
-  },
+  }
 
-  init: function (config, callback) {
+  self.init = function (config, callback) {
```
masylum: and here comes the `self` :smile:

diasdavid: ahah eventually we will run perf analysis everywhere and there will be a cascade of .prototype and named functions, until then, this makes the code so easy to read :D

PR #16, opened 2015-12-15
feature complete + tests

diasdavid:


jbenet:
Awesome!!!! :)

src/stores/config.js
```diff
@@ -1,19 +1,33 @@
...
-      return store.read('config', function (err, content) {
-        if (err) return cb(err)
+    get: function (callback) {
+      store
+        .createReadStream('config')
+        .pipe(bl(function (err, config) {
+          if (err) {
+            return callback(err)
+          }
+          callback(null, JSON.parse(config))
```
masylum: `JSON.parse` can raise an exception, this is why I had the try/catch

diasdavid: good catch! :) fixed here https://github.com/ipfs/js-ipfs-repo/pull/18

PR #19, opened 2015-12-17
Fixes

ralphtheninja:


src/stores/config.js
```diff
@@ -12,11 +12,13 @@ exports.setUp = function (basePath, blobStore, locks) {
           if (err) {
             return callback(err)
           }
+          var result
           try {
-            callback(null, JSON.parse(config))
```
diasdavid: Why doesn't this work?

diasdavid: If JSON.parse throws, the error is automatically catch by the try/catch clause, here is a simplier example:

```JavaScript
try {
  checkIfRuns(JSON.parse('sdadadas:asda'))
} catch (e) {
  console.log('=> error')
}

function checkIfRuns (arg1) {
  console.log('->', arg1)
}
```

```bash
$ node try-catch.js
=> error
```

ralphtheninja: It works but if *callback* throws that will also be catched.

diasdavid: I see, good catch! thank you @ralphtheninja :)

ralphtheninja: My first code contribution. Yay! :dancers:

jbenet: :fireworks: :tada: :smile:

PR #26, opened 2016-01-19
Running the tests in the browser

xicombd:
Hey!

I'm trying to get the repo running on a browser, but I'm having some trouble doing it, so a review would be appreciated :)

I started by specifying two different entry points on `package.json`, one for now which uses `fs-blob-store` as before and one for the browser which uses `idb-blob-store`.

I've also started setting up karma for running the tests, however they're currently very slow and inconsistent, having different results on different runs, but I haven't really understood why.

Please let me know if you find anything


diasdavid:
Let's finish this up today after ArcticJS talks :)

tests/browser-tests.js
```diff
@@ -0,0 +1,61 @@
...
+      const blob = _.startsWith(file.key, 'blocks/')
+        ? blocksBlob
+        : mainBlob
+
+      blob.createWriteStream({
+        key: file.key
+      }).end(file.value, cb)
+    }, () => {
+      // FIXME debuging possible idb-blob-store bug
+      setTimeout(done, 1000)
```
diasdavid: This line was added for testing purposes.

It seems that the `end` event on the returned write streams is emitted before it actually finishes, making the testes fail, since the db isn't consistent.

tests/repo-test.js
```diff
@@ -71,127 +26,148 @@ describe('IPFS Repo Tests', () => {
...
+          return done()
+        }
 
-    it('block exists', done => {
-      const buf = new Buffer(base58.decode(baseFileHash))
+        const buf = new Buffer(base58.decode(baseFileHash))
+        repo.datastore.createReadStream(buf)
+          .pipe(bl((err, data) => {
+            expect(err).to.not.exist
+            const eq = fs.readFileSync(process.cwd() + '/tests/test-repo/blocks/12207028/122070286b9afa6620a66f715c7020d68af3d10e1a497971629c07606bfdb812303d.data').equals(data)
```
diasdavid: Can you `brfs` this file so that the test also works in the browser?

tests/browser-tests.js
diasdavid: Is this still required with `local-storage-blob-store`?

tests/repo-test.js
```diff
@@ -71,127 +26,148 @@ describe('IPFS Repo Tests', () => {
...
 
-    it('set config', done => {
-      repo.config.set({a: 'b'}, err => {
-        expect(err).to.not.exist
+    describe('config', () => {
+      it('get config', done => {
         repo.config.get((err, config) => {
           expect(err).to.not.exist
-          expect(config).to.deep.equal({a: 'b'})
+          expect(config).to.be.a('object')
```
diasdavid: why the change on this assertion?

tests/browser-tests.js
xicombd: Nope, just removed the `setTimeout`

tests/repo-test.js
xicombd: I didn't change that, github seems to be mixing different tests

diasdavid: got it, thanks :)

karma.conf.js
```diff
@@ -0,0 +1,80 @@
...
+
+    // If browser does not capture in given timeout [ms], kill it
+    captureTimeout: 60000,
+
+    // Continuous Integration mode
+    // if true, it capture browsers, run tests and exit
+    singleRun: true,
+
+    // List plugins explicitly, since autoloading karma-webpack
+    // won't work here
```
Dignifiedquire: why? this should work if they are listed in package.json

tests/browser-tests.js
```diff
@@ -0,0 +1,57 @@
+/* globals describe, before */
+
+// const expect = require('chai').expect
+const async = require('async')
+const store = require('local-storage-blob-store')
+const tests = require('./repo-test')
+const _ = require('lodash')
+const IPFSRepo = require('../src')
+
+global.Buffer = require('buffer').Buffer
```
Dignifiedquire: why make this global? you can shim it with webpack

tests/browser-tests.js
```diff
@@ -0,0 +1,57 @@
...
+
+var repoContext = require.context('raw!./test-repo', true)
```
Dignifiedquire: why require.context?

Dignifiedquire:
I can try running it later.  please let me know if you made progress so I don't do things twice

diasdavid:
@xicombd, can you check @Dignifiedquire comments and adjust to those?

Dignifiedquire:
Cleared all my confusions and running well on my machine, just left some notes about things that could be improved

xicombd:
> @xicombd, can you check @Dignifiedquire comments and adjust to those?
Yep, on it right now.

@Dignifiedquire everything is working now, the problems I was having were apparently related to [`idb-blob-store`](https://github.com/substack/idb-blob-store) racing conditions, so I ended up implementing [´local-storage-blob-store´](https://github.com/xicombd/local-storage-blob-store) which seems to be doing it's job.

tests/browser-tests.js
xicombd: only option that I remembered, how would you do that?

karma.conf.js
xicombd: first time using karma, just removed that :)

tests/browser-tests.js
Dignifiedquire: Oh wait this is recursively pulling in a folder? sorry didn't realise it, then :+1:

xicombd: Yeah, it's loading all the repo files so they can be inserted later on localStorage

tests/browser-tests.js
xicombd: I'm not being able to find a way of doing this :( 

We need is `require('buffer').Buffer` and not just `require('buffer')`, and it seems that [`webpack.ProvidePlugin`](https://webpack.github.io/docs/shimming-modules.html#plugin-provideplugin) can't do it. 

Do you know any other way?

Dignifiedquire: Take a look [here](https://webpack.github.io/docs/configuration.html#node)

Adding `node.Buffer = true` should do the trick

xicombd: Ahh, that's awesome, thanks for the tip :D

diasdavid:
🎉🎉🎉🎉
:D 

Thank you for finishing this @xicombd :)

PR #27, opened 2016-01-27
Get tests to pass using indexeddb instead of local-storage

Dignifiedquire:
## Issues

- [x] Writes are not flushed when the write stream `finish`es. Fixed in idb-plus-blob-store
- [x] At the moment reads are failing because of https://github.com/feross/typedarray-to-buffer/issues/5 Fixed using https://github.com/webpack/webpack/pull/1931
- [x] Cleanup handling of pathing issues for the store names

### DIFF
--- a/src/stores/config.js
+++ b/src/stores/config.js
@@ -14,7 +14,7 @@ exports.setUp = (basePath, blobStore, locks) => {
           }
           var result
           try {
-            result = JSON.parse(config)
+            result = JSON.parse(config.toString())
           } catch (err) {
             return callback(err)
           }
--- a/tests/browser-tests.js
+++ b/tests/browser-tests.js
@@ -1,18 +1,23 @@
 /* globals describe, before */
 
-// const expect = require('chai').expect
 const async = require('async')
-const store = require('local-storage-blob-store')
+const store = require('idb-plus-blob-store')
 const tests = require('./repo-test')
 const _ = require('lodash')
 const IPFSRepo = require('../src')
 
-var repoContext = require.context('raw!./test-repo', true)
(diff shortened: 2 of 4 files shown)


src/stores/config.js
```diff
@@ -14,7 +14,7 @@ exports.setUp = (basePath, blobStore, locks) => {
           }
           var result
           try {
-            result = JSON.parse(config)
+            result = JSON.parse(config.toString())
```
diasdavid: didn't this ever worked?

Dignifiedquire: No idea how this ever worked before

Dignifiedquire: Actually I know, local storage converts everything to strings, as it can't store blobs, so that's why that worked is my guess

tests/browser-tests.js
```diff
@@ -1,18 +1,23 @@
...
+        window.mozIndexedDB ||
+        window.webkitIndexedDB ||
+        window.msIndexedDB
+
+idb.deleteDatabase('ipfs')
+idb.deleteDatabase('ipfs/blocks')
 
+describe('IPFS Repo Tests on the Browser', function () {
+  before(function (done) {
     var repoData = []
```
diasdavid: do we need `var` here?

Dignifiedquire: I didn't touch that line

Dignifiedquire: So no idea why there are `var`s in some places

diasdavid:
This is pretty awesome @Dignifiedquire , thank you for fixing it so quickly :)

src/stores/config.js
diasdavid: :+1:

PR #28, opened 2016-01-28
Use buffer-loader on tests instead of raw-loader

xicombd:
Makes file buffers have the same content and length on both node and browser. 

Related to https://github.com/ipfs/js-ipfs/pull/62 and https://github.com/ipfs/js-ipfs/pull/61


dignifiedquire:
LGTM

PR #29, opened 2016-03-16
fix: Upgrade dependencies, use strict and fix /blocks pathing

dignifiedquire:
For @nginnever :)


diasdavid:
what was the pathing issue?

src/stores/datastore.js
```diff
@@ -1,35 +1,37 @@
+'use strict'
+
 const PREFIX_LENGTH = 8
 
 exports = module.exports
 
 exports.setUp = (basePath, blobStore, locks) => {
-  var store = blobStore(basePath + '/blocks')
+  const store = blobStore(basePath + '/blocks')
```
dignifiedquire: It was here, in the version that was installed from npm, looks like I didn't fix it but the fix is not yet released to npm

dignifiedquire: it said before `/blocks/`

diasdavid: I'm confused

diasdavid:
@dignifiedquire I'm still unaware of what was the bug this PR is fixing and what is the actual solution. Maybe @nginnever knows the answer?

dignifiedquire:
Sorry will explain tomorrow

dignifiedquire:
@diasdavid the most important fix is adding  `'use strict'` to all files, as the combination of non strict mode with some es6 features like `const` inside some browsers resulted in unpredictable behaviour when not transpiled through babel. (Remember those random tests failing for @nginnever, that was the reason for that)

The other issue was that in an older revision the repo path for the blocks was `/blocks/` but in newer revisions it is `/blocks`. But that was fixed before this PR, but wasn't published to npm.

diasdavid:
Thank you @dignifiedquire . Also good discussion on the call today @dignifiedquire @nginnever  :)

PR #30, opened 2016-03-17
No longer considers a non-existant repo an error.

noffle:
Calling repo.exists() on a repo that doesn't exist would return an error
rather than false, which would also mask whatever the underlying error
really was.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,21 +4,24 @@ exports = module.exports = Repo
 
 function Repo (repoPath, options) {
   this.init = (config, callback) => {
-    if (this.exists()) {
-      throw new Error('Repo already exists')
-    }
+    this.exists((err, exists) => {
+      if (err) { throw err }
+      if (exists) { throw new Error('Repo already exists') }
+      throw new Error('not implemented')
+    })
   }
 
   this.locks = stores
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -4,21 +4,24 @@ exports = module.exports = Repo
 
 function Repo (repoPath, options) {
   this.init = (config, callback) => {
-    if (this.exists()) {
-      throw new Error('Repo already exists')
-    }
+    this.exists((err, exists) => {
+      if (err) { throw err }
+      if (exists) { throw new Error('Repo already exists') }
+      throw new Error('not implemented')
```
diasdavid: 'not implemented' or 'repo not found'?

diasdavid:
sweet, good catch! :) thank you noffle. Just made a tiny comment

src/index.js
noffle: There's no further implementation of `repo.init` yet, so 'not implemented' seemed fitting.

noffle:
Ah, sorry, just saw your comments :)

src/index.js
diasdavid: got it! :)

PR #32, opened 2016-03-22
Use dignified.js

dignifiedquire:
Ref https://github.com/dignifiedquire/dignified.js/issues/1


diasdavid:
let's do https://github.com/ipfs/js-ipfs-repo/pull/36 since you already added dignified there too :)

PR #44, opened 2016-04-20
fix: Ensure callbacks are only called once

dignifiedquire:


src/stores/config.js
```diff
@@ -31,7 +31,7 @@ exports.setUp = (basePath, blobStore, locks) => {
         }
 
         store.createWriteStream('config')
-          .on('finish', () => {
+          .once('finish', () => {
```
diasdavid: there was several finish events? wat?

dignifiedquire: yep :cry: making my debugging life very sad

PR #54, opened 2016-04-27
fix extension

diasdavid:
found that extensions were defaulting to 'data', while running block-service tests. This PR fixes it


diasdavid:
@dignifiedquire

dignifiedquire:
Please add a test for this

diasdavid:
@dignifiedquire there are already tests for custom extensions, but since it was always defaulting to 'data' and the block was the same `hello world`, it was able to work flawlessly

diasdavid:
@dignifiedquire updated the tests to create a different object `hello world 2` for the custom extension, so that the error can't be missed again.

dignifiedquire:
lgtm

PR #59, opened 2016-05-05
fix: follow abstract-blob-store interface for exists calls

dignifiedquire:


src/index.js
```diff
@@ -30,13 +30,7 @@ function Repo (repoPath, options) {
...
 
   this.exists = (callback) => {
-    this.version.exists((err, exists) => {
-      if (err) {
-        callback(new Error('repo does not exist'), false)
-      } else {
-        callback(null, exists)
-      }
-    })
+    this.version.exists(callback)
```
diasdavid: perhaps `this.exists = this.version.exists.bind(this)` ?

dignifiedquire: or we could do `this.exists = (cb) => this.version.exists(cb)` ;) 

Not sure tbh if there is a big difference between those but can do

PR #85, opened 2016-09-12
fix(blockstore: lock getStream to avoid race issues

dignifiedquire:
Actually fixes the issue in https://github.com/ipfs/js-ipfs/pull/469


test/blockstore-test.js
```diff
@@ -109,7 +109,7 @@ module.exports = (repo) => {
 
       it('massive read', (done) => {
         parallel(_.range(20 * 100).map((i) => (cb) => {
-          const j = i % 20
+          const j = i % 50
```
VictorBjelkholm: Can we put this magic number in a variable that explains what it means?

dignifiedquire: fixed

test/blockstore-test.js
```diff
@@ -109,7 +109,7 @@ module.exports = (repo) => {
 
       it('massive read', (done) => {
         parallel(_.range(20 * 100).map((i) => (cb) => {
-          const j = i % 20
+          const j = i % blockCollection.length
```
diasdavid: What's this value?

src/stores/blockstore.js
```diff
@@ -52,19 +52,22 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
+        pull(
+          store.read(p),
+          pull.collect(release((err, data) => {
+            if (err) return deferred.abort(err)
+            deferred.resolve(pull.values([
+              new Block(Buffer.concat(data), ext)
+            ]))
+          }))
+        )
+      })
```
diasdavid: Why do streams -> buffer everything -> streams?

src/stores/blockstore.js
```diff
@@ -52,19 +52,22 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
 
       const p = multihashToPath(key, extension)
+      const deferred = defer()
+
+      lock(p, (release) => {
+        const ext = extension === 'data' ? 'protobuf' : extension
+        pull(
+          store.read(p),
+          pull.collect(release((err, data) => {
+            if (err) return deferred.abort(err)
```
diasdavid: can we keep consistency with the `curlys`, please

src/stores/blockstore.js
diasdavid: update: nvm. It is because of `interface-pull-blob-store` only streams interface. We should consider to have a sync interface too on the blob-store.

test/blockstore-test.js
dignifiedquire: What value? This is iterating over the earlier put `blockCollection` doing parallel reads

src/stores/blockstore.js
dignifiedquire: consistent with no curlys? :)

diasdavid: I do remember having a chat about this a long time ago and we agreed with curlys, AFAIK that decision hasn't changed

dignifiedquire: hmmmmmmmmmmmmmm <insert snarky comment>

dignifiedquire: there you go

diasdavid:
thank you @dignifiedquire  :)

PR #89, opened 2016-10-03
Awesome IPLD Endeavour

diasdavid:
- [x] remove extension support (no need with CID)
- [x] blockstore starts getting blockBlobs instead of blocks (so that it can know the right key)

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   "dependencies": {
     "babel-runtime": "^6.11.6",
     "base32.js": "^0.1.0",
-    "ipfs-block": "^0.3.0",
+    "ipfs-block": "^0.4.0",
     "lock": "^0.1.3",
     "multihashes": "^0.2.2",
     "pull-defer": "^0.2.2",
--- a/src/stores/blockstore.js
+++ b/src/stores/blockstore.js
@@ -5,16 +5,16 @@ const pull = require('pull-stream')
 const Lock = require('lock')
 const base32 = require('base32.js')
 const path = require('path')
-const write = require('pull-write')
+const pullWrite = require('pull-write')
 const parallel = require('run-parallel')
-const defer = require('pull-defer/source')
+const pullDefer = require('pull-defer/source')
 
 const PREFIX_LENGTH = 5
 
 exports = module.exports
 
-function multihashToPath (multihash, extension) {
-  extension = extension || 'data'
+function multihashToPath (multihash) {
+  const extension = 'data'
   const encoder = new base32.Encoder()
   const hash = encoder.write(multihash).finalize()
   const filename = `${hash}.${extension}`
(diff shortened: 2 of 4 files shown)


dignifiedquire:
Why not add cid support directly to Block, The current `{block, cid}` constructs seems akward and easy to break accidentally.

diasdavid:
I scratched my head a lot thinking about this. Although having a .cid() on the block would be super sweet, you can't really tell which hash function to use, otherwise, it breaks the Block abstraction.

Only the Block Service, hinted by the caller with the CID, can tell which data type and hash function should it use for the data it is looking for.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ipfs-repo",
   "version": "0.9.1",
   "description": "IPFS Repo implementation",
-  "main": "lib/index.js",
+  "main": "src/index.js",
```
dignifiedquire: TODO: change before merge

src/stores/blockstore.js
```diff
@@ -5,16 +5,16 @@ const pull = require('pull-stream')
 const Lock = require('lock')
 const base32 = require('base32.js')
 const path = require('path')
-const write = require('pull-write')
+const pullWrite = require('pull-write')
```
dignifiedquire: why?

src/stores/blockstore.js
```diff
@@ -5,16 +5,16 @@ const pull = require('pull-stream')
...
+const pullDefer = require('pull-defer/source')
 
 const PREFIX_LENGTH = 5
 
 exports = module.exports
 
-function multihashToPath (multihash, extension) {
-  extension = extension || 'data'
+function multihashToPath (multihash) {
+  const extension = 'data'
```
dignifiedquire: if this is fixed, why not make it a constant like `PREFIX_LENGTH`

src/stores/blockstore.js
```diff
@@ -27,82 +27,113 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
-
-            written.push(meta)
-            cb()
-          })
-        }), cb)
+      const sink = pullWrite((blockBlobs, cb) => {
+        const tasks = blockBlobs.map((blockBlob) => {
+          return (cb) => {
+            writeBlock(blockBlob, (err, meta) => {
+              if (err) {
```
dignifiedquire: the nesting is strong in this one,  maybe simplify this somehow?

src/stores/locks.js
```diff
@@ -37,16 +41,22 @@ exports.setUp = (basePath, BlobStore) => {
...
+              return cb(err)
+            }
+
+            if (exists) {
+              return cb(new Error('failed to remove lock'))
+            }
+
+            cb()
+          })
+        }
```
dignifiedquire: :-1: this change makes it harder to read, and is a matter of personal preference

src/stores/blockstore.js
```diff
@@ -27,82 +27,113 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
+        parallel(tasks, cb)
       }, null, 100, (err) => {
         ended = err || true
-        if (push) push(ended)
+        if (push) {
+          push(ended)
+        }
       })
 
+      // TODO ??Why does a putStream need to be a source as well??
```
diasdavid: @dignifiedquire could you clarify on this one?

src/stores/blockstore.js
```diff
@@ -27,82 +27,94 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
+        parallel(tasks, cb)
       }, null, 100, (err) => {
         ended = err || true
-        if (push) push(ended)
+        if (push) {
+          push(ended)
+        }
       })
 
+      // TODO ?? Why does a putStream need to be a source as well??
```
diasdavid: @dignifiedquire

diasdavid:
for record, @dignifiedquire approved

PR #105, opened 2016-11-30
Fix js code sample

hoffmabc:


diasdavid:
thank you @hoffmabc :)

hoffmabc:
Always making big impacts

README.md
```diff
@@ -120,9 +120,9 @@ Example:
 
 ```js
 var fsBlobStore = require('fs-blob-store')  // an in-memory blob store
```
dignifiedquire: actually this is not correct anymore, it needs to be `fs-pull-blob-store` and the comment is wrong as this is not an in-memory blob store, but an fs blob store

dignifiedquire: cc @diasdavid

diasdavid: true, should had tested it. Fixed it now https://github.com/ipfs/js-ipfs-repo/commit/c2c76d7df55381e89f3a3558b9f98962e3a1dd46

PR #120, opened 2017-02-14
blockstore - allow for multiple backing stores

kumavis:
This feature allows the blockstore to support multiple blob stores simultaneously. All stores are read from and written to in parallel, but when reading we take the fastest non-error result.

**In order to effectively test** both simple and advanced **configurations against** node + browser **platforms**, I needed to make some changes to how the tests are run. **Tests are now atomic** and do not rely on any side effects from previous tests. This incurs a slow down. `npm test` (linting + node, browser, browser-webworker environment tests) takes 1.77 min on my laptop
 Instead of passing in the `repo` at test definition time we pass in a `testConfig` object that gets a `repo` instantiated and attached before each test.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -41,16 +41,12 @@ module.exports = class Repo {
 }
 
 function initializeBlobStores (store) {
-  if (store.constructor) {
-    return {
-      keys: store,
-      config: store,
-      blockstore: store,
-      logs: store,
-      locks: store,
-      version: store
-    }
+  return {
+    keys: store.keys || store,
+    config: store.config || store,
+    blockstore: store.blockstore || store,
+    logs: store.logs || store,
+    locks: store.locks || store,
+    version: store.version || store
   }
-
-  return store
 }
--- a/src/stores/blockstore.js
+++ b/src/stores/blockstore.js
@@ -6,8 +6,9 @@ const base32 = require('base32.js')
 const path = require('path')
 const pull = require('pull-stream')
 const pullWrite = require('pull-write')
-const pullDefer = require('pull-defer/source')
+const pullDefer = require('pull-defer')
 const parallel = require('async/parallel')
+const each = require('async/each')
 
 const PREFIX_LENGTH = 5
 const EXTENSION = 'data'
@@ -24,7 +25,10 @@ function multihashToPath (multihash) {
 }
 
 exports.setUp = (basePath, BlobStore, locks) => {
-  const store = new BlobStore(basePath + '/blocks')
+  const BlobStores = Array.isArray(BlobStore) ? BlobStore : [BlobStore]
+  const stores = BlobStores.map((BlobStoreClass) => {
+    return new BlobStoreClass(basePath + '/blocks')
+  })
   const lock = new Lock()
 
   // blockBlob is an object with:
@@ -37,12 +41,15 @@ exports.setUp = (basePath, BlobStore, locks) => {
     const key = multihashToPath(blockBlob.key)
 
     lock(key, (release) => {
-      pull(
-        pull.values([
-          blockBlob.data
-        ]),
-        store.write(key, release(released))
-      )
+      // write to each store in parallel
+      each(stores, (store, cb) => {
+        pull(
+          pull.values([
+            blockBlob.data
+          ]),
+          store.write(key, cb)
+        )
+      }, release(released))
     })
 
     // called once the lock is released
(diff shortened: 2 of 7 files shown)


src/stores/blockstore.js
```diff
@@ -37,12 +41,15 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
-      )
+      // write to each store in parallel
+      each(stores, (store, cb) => {
+        pull(
+          pull.values([
+            blockBlob.data
+          ]),
+          store.write(key, cb)
+        )
+      }, release(released))
```
dignifiedquire: The locking should happen on a per store basis.

src/index.js
```diff
@@ -41,16 +41,12 @@ module.exports = class Repo {
...
+  return {
+    keys: store.keys || store,
+    config: store.config || store,
+    blockstore: store.blockstore || store,
+    logs: store.logs || store,
+    locks: store.locks || store,
+    version: store.version || store
   }
-
-  return store
```
dignifiedquire: why this change?

src/stores/blockstore.js
```diff
@@ -62,28 +69,50 @@ exports.setUp = (basePath, BlobStore, locks) => {
...
 
+      // lock, read from all stores in parallel, take first returned result, and release
       lock(blockPath, (release) => {
-        pull(
-          store.read(blockPath),
-          pull.collect(release(released))
-        )
+        const onComplete = release(onReleased)
+        let didComplete = false
+        each(stores, (store, cb) => {
```
dignifiedquire: Something like `race` would make more sense here, as you only care for the first one to return.

src/stores/blockstore.js
```diff
@@ -62,28 +69,50 @@ exports.setUp = (basePath, BlobStore, locks) => {
       }
 
       const blockPath = multihashToPath(key)
-      const deferred = pullDefer()
+      const deferredSource = pullDefer.source()
+      // const deferredSource = pullDefer.source()
 
+      // lock, read from all stores in parallel, take first returned result, and release
```
dignifiedquire: again, locking should happen on a per store basis

src/stores/blockstore.js
```diff
@@ -175,7 +204,15 @@ exports.setUp = (basePath, BlobStore, locks) => {
       }
 
       const blockPath = multihashToPath(key)
-      store.exists(blockPath, callback)
+
+      // check all stores for exists in parallel
+      parallel(stores.map((store) => {
```
dignifiedquire: `race` would make more sense here, as you are only interested in one result

kumavis:
hey back from vaca-- picking this up again

src/stores/blockstore.js
kumavis: I used `race` on my first pass, but found it easier to use parallel to get the correct behavior in all cases

`race` did not work well bc we wanted to return on first non-error result

src/stores/blockstore.js
kumavis: same reasoning as https://github.com/ipfs/js-ipfs-repo/pull/120/files#r103351131

src/index.js
kumavis: slightly out of bound, but this was a confusing dead code path.
`store.constructor` is always truthy, even for `{}`

src/stores/blockstore.js
kumavis: if this is true then it seems like the locking should be implemented at the blobStore level

dignifiedquire:
Hey @kumavis I have been thinking about this a bit and I think a better solution than putting this code into js-ipfs-repo would be to write a new pull-blob-store which has the same interface as the others, but is a combiner, so you could use it like this:


```js
const multi = new MultiStore([store1, store2, store3])
const repo = new Repo('my-repo, {stores: multi})
```

what do you think?

kumavis:
yeah that might make more sense but maybe there is some edgecase... ill think about it. does seem cleaner.

dignifiedquire:
@kumavis this is in reference to https://github.com/ipfs/go-datastore/blob/master/tiered/tiered.go which is the equivalent on the go side

kumavis:
ooo

dignifiedquire:
@kumavis if you like that, take a look at https://github.com/dignifiedquire/js-datastore. I've just finished implementing the tiered one for js. This will be integrated into ipfs-repo soon

kumavis:
@dignifiedquire ok cool - should i close this now?
@diasdavid any reason to have native multi-store support? thoughts on @dignifiedquire 's approach?

diasdavid:
@kumavis multistore support is now 'native' thanks to datastore approach of 'mounting'. So yeah, you won't needs this PR anymore. We should have planned better so that you wouldn't waste time with this.

PR #123, opened 2017-03-14
Datastore integration

dignifiedquire:
Dragons inside ! 🐉 🐉 🐉 

Integration with the new datastores, will have breaking api changes! 
Will fix https://github.com/ipfs/js-ipfs/issues/229 due to using https://github.com/dignifiedquire/lock-me for aquiring file system locks and exposing need methods to read and write the http api addr file.

ref https://github.com/ipfs/js-ipfs/issues/787

- [x] locking working in browser and node
- [x] expose new init function
- [x] disable flatfs sharding in the browser
- [x] flatfs compatible with the latest go-ipfs
- [x] blockstore interop with go
- [x] read and write config
- [x] read and write repo version
- [x] datastore interop 
- [x] datastore tests


src/index.js
```diff
@@ -1,56 +1,246 @@
...
+  /**
+   * Close the repo and cleanup.
+   *
+   * @param {function(Error)} callback
+   * @returns {void}
+   */
+  close (callback) {
+    if (this.closed) {
+      return callback(new Error('repo is already closed'))
     }
```
diasdavid: Different behaviour. open will return silently if it is already open, but close will return an error. Or return an error in both or silent in both.

src/index.js
```diff
@@ -1,56 +1,246 @@
...
+   * @returns {void}
+   */
+  getPrivateKey (callback) {
+    this.config.get((err, config) => {
+      if (err) {
+        return callback(err)
+      }
+      callback(null, config.Identity.PrivKey)
+    })
+  }
```
diasdavid: Why do we need this method here? Repo should not assume a specific format of the config file.

README.md
```diff
@@ -45,62 +34,78 @@ This is the implementation of the [IPFS repo spec](https://github.com/ipfs/specs
...
-more. Each of the individual repos has an interface defined by
-[interface-pull-blob-store](https://github.com/ipfs/interface-pull-blob-store): this
-enables us to make IPFS Repo portable (running on Node.js vs the browser) and
-accept different types of storage mechanisms for each repo (fs, levelDB, etc).
-
-### Good to know (historical context)
-
-- The datastore folder holds the legacy version of datastore, still built in levelDB, there is a current endeavour of pushing it to fs completely.
-- The blocks folder is the current version of datastore.
-- The keys repo doesn't exist yet, as the private key is simply stored inside config
```
dignifiedquire: It's not correct anymore, that's why I removed it for now

example.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
```
dignifiedquire: no, this is used in the docs generation `example.js` is a special file name

package.json
```diff
@@ -3,24 +3,31 @@
   "version": "0.11.3",
   "description": "IPFS Repo implementation",
   "main": "src/index.js",
+  "browser": {
+    "./src/lock.js": "./src/lock-browser.js",
+    "./src/default-config.js": "./src/default-config-browser.js"
+  },
```
dignifiedquire: fair point, any thoughts on better naming?

src/index.js
dignifiedquire: This was in here before, I just migrated the functionality

README.md
```diff
@@ -45,62 +34,78 @@ This is the implementation of the [IPFS repo spec](https://github.com/ipfs/specs
...
+         ┌───────────┴───────────┐                           ┌───────────┴───────────┐
+         │                       │                           │                       │
+         │                       │                           │                       │
+┌─────────────────┐     ┌─────────────────┐         ┌─────────────────┐     ┌─────────────────┐
+│                 │     │                 │         │                 │     │                 │
+│     /blocks     │     │   /datastore    │         │     /blocks     │     │   /datastore    │
+│                 │     │                 │         │                 │     │                 │
+├─────────────────┤     ├─────────────────┤         ├─────────────────┤     ├─────────────────┤
+│ FlatfsDatastore │     │LevelDBDatastore │         │LevelJSDatastore │     │LevelJSDatastore │
+└─────────────────┘     └─────────────────┘         └─────────────────┘     └─────────────────┘
```
dignifiedquire: it's a drawing of the datastores and how the are used, what you had before were not actual stores

src/index.js
diasdavid: ␡💥

package.json
```diff
@@ -31,27 +38,29 @@
     "node": ">=4.0.0"
```
dignifiedquire: done

src/index.js
dignifiedquire: done, will always error now

src/index.js
dignifiedquire: done

package.json
dignifiedquire: renamed to `default-options`

README.md
```diff
@@ -119,77 +124,26 @@ Loading this module through a script tag will make the `IpfsRepo` obj available
...
-var inMemoryBS = require('interface-pull-blob-store')
-// inMemoryBS is an "in memory" blob store, you can find others at:
-// https://github.com/ipfs/interface-pull-blob-store#modules-that-use-this
-
-var IPFSRepo = require('ipfs-repo')
-var repo = new IPFSRepo('/Users/someone/.ipfs', {
-  stores: inMemoryBS
+const Repo = require('ipfs-repo')
+const repo = new Repo('/Users/awesome/.jsipfs')
+repo.init(defaultConfig, (err) => {
```
dignifiedquire: fixed

README.md
```diff
@@ -119,77 +124,26 @@ Loading this module through a script tag will make the `IpfsRepo` obj available
...
 ```js
-var inMemoryBS = require('interface-pull-blob-store')
-// inMemoryBS is an "in memory" blob store, you can find others at:
-// https://github.com/ipfs/interface-pull-blob-store#modules-that-use-this
-
-var IPFSRepo = require('ipfs-repo')
-var repo = new IPFSRepo('/Users/someone/.ipfs', {
-  stores: inMemoryBS
+const Repo = require('ipfs-repo')
+const repo = new Repo('/Users/awesome/.jsipfs')
```
dignifiedquire: done

README.md
diasdavid: Only the last item is not valid anymore

example.js
diasdavid: I see. I'm not very keen of putting even more stuff at the root level of each repo.

README.md
```diff
@@ -119,77 +124,26 @@ Loading this module through a script tag will make the `IpfsRepo` obj available
...
+    throw err
+  }
+
+  repo.open((err) => {
+    if (err) {
+      throw err
+    }
+
+    console.log('repo is ready')
+  })
```
diasdavid: 🛎

example.js
dignifiedquire: we already have it in many repos, so what do you want to do about it?

README.md
dignifiedquire: the first item is also not true, datastore is not legacy, it's being actively used and things added to it

dignifiedquire: which makes the second incorrect as well

dignifiedquire: making the last item the only one valid

README.md
dignifiedquire: done

example.js
diasdavid: Nothing now, let's revisit docs strategy and execution for next quarter

package.json
diasdavid: thank you :)

Issue #127, opened 2017-03-22
Investigate level-js status


dignifiedquire:
The current master of https://github.com/maxogden/level.js is not up to date with the rest of the level eco system. Currently we are depending on https://github.com/timkuijsten/level.js/tree/idbunwrapper which is near complete rewrite supporting use cases like empty values for us. 

We should try and see  in https://github.com/maxogden/level.js/issues/46 what it takes to finish this and get it released or start maintaining our own version.

diasdavid:
@pgte you are very close to the Level ecosystem. Any comment here?

pgte:
I'm less familiar with the level-js than the node side of it..
Anyway, I've been implementing a level down store myself, and the abstract level down test suite is impressive, so maintaining such a thing is not something to be taken light-heartedly, specially if this is touching the (de)serialising / (de)encoding parts.. :)
Which leads me to asking:
What problem are we trying to prevent / solve here?

dignifiedquire:
The issue is that the original level.js is not usable for us which is why we are currently using a fork, which brings with it all the issues of depending on a fork which is not maintained nor officially released. 

Main issues with level.js master were the iterators not being usable as far as I remember and some other thing that are fixed in the fork

diasdavid:
@dignifiedquire is there a PR to the main one?

dignifiedquire:
@diasdavid no just this comment:https://github.com/maxogden/level.js/issues/46#issuecomment-182144244

ralphtheninja:
@dignifiedquire o/ I have begun working on this.

ralphtheninja:
Quick and dirty roadmap:

* move `level.js` to `level` org, me, @juliangruber and @vweevers are active there now, I have asked @maxogden to move it and he's cool with it
* spend some time on fixing most obvious problems to get it up to par with `abstract-leveldown`
* replace the `idb-wrapper`
* get snapshots working

I haven't looked at the fork just yet, just focusing on the big brush for now :)

dignifiedquire:
@ralphtheninja great to hear, the fork works very well and is also a good deal faster than the official version if I remember correctly. I also think the other reason to use the fork were issues with the iterators, but I don't remember the details atm.

juliangruber:
I'm happy with what @ralphtheninja suggests!

ralphtheninja:
[...]

PR #128, opened 2017-03-22
fix: use open method and fork of level-js

dignifiedquire:
ref https://github.com/ipfs/js-ipfs-repo/issues/127

depends on https://github.com/ipfs/interface-datastore/pull/4


src/index.js
```diff
@@ -192,10 +206,15 @@ class IpfsRepo {
...
+        log('unlocking')
         this.closed = true
         this.lockfile.close(cb)
+      },
+      (cb) => {
+        this.lockfile = null
+        cb()
       }
-    ], callback)
+    ], (err) => callback(err))
```
diasdavid: pass the callback directly.

PR #136, opened 2017-06-04
Rename stores

justinmchase:
I'm intending to add a new abstraction for s3 file storage (#135). This is simply a straight up refactor to rename the stores based on their purpose rather than the names of the kinds of storage. In other words, instead of `fs` and `level` the stores are renamed to `block` and `data`.

The s3 storage type will be in another repo and will be very similar to the fs datastore but saving into s3.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -199,9 +199,8 @@ class IpfsRepo {
         cb(err)
       }),
       (cb) => this.store.close(cb),
-      (cb) => console.log('closing block...') || this._blockStore.close(cb),
+      (cb) => this._blockStore.close(cb),
       (cb) => {
-        console.log('lock:', this.lockfile)
         log('unlocking')
         this.closed = true
         this.lockfile.close(cb)
--- a/src/index.js
+++ b/src/index.js
@@ -127,13 +127,16 @@ class IpfsRepo {
         const DataStore = this.options.dataStore
         const dataStore = new DataStore(path.join(this.path, dataStoreDirectory), this.options.dataStoreOptions)
         log(dataStore)
-        this.store = new MountStore([{
-          prefix: new Key(blockStoreDirectory),
-          datastore: blockStore
-        }, {
-          prefix: new Key('/'),
-          datastore: dataStore
-        }])
+        this.store = new MountStore([
+          {
+            prefix: new Key(blockStoreDirectory),
+            datastore: blockStore
+          },
+          {
+            prefix: new Key('/'),
+            datastore: dataStore
+          }
+        ])
 
         this.blockstore = blockstore(this)
         this.closed = false
@@ -196,8 +199,9 @@ class IpfsRepo {
         cb(err)
       }),
       (cb) => this.store.close(cb),
-      (cb) => this._blockStore.close(cb),
+      (cb) => console.log('closing block...') || this._blockStore.close(cb),
       (cb) => {
+        console.log('lock:', this.lockfile)
         log('unlocking')
         this.closed = true
         this.lockfile.close(cb)
--- a/src/index.js
+++ b/src/index.js
@@ -5,7 +5,6 @@ const MountStore = core.MountDatastore
 const ShardingStore = core.ShardingDatastore
 
 const Key = require('interface-datastore').Key
-const LevelStore = require('datastore-level')
 const waterfall = require('async/waterfall')
 const series = require('async/series')
 const parallel = require('async/parallel')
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -131,17 +122,21 @@ class IpfsRepo {
...
-          datastore: flatfs
-        }, {
-          prefix: new Key('/'),
-          datastore: new LevelStore(path.join(this.path, levelDirectory), {
-            db: this.options.level
-          })
-        }])
+        const DataStore = this.options.dataStore
+        const dataStore = new DataStore(path.join(this.path, dataStoreDirectory), this.options.dataStoreOptions)
+        log(dataStore)
```
dignifiedquire: this should be removed

test/node.js
```diff
@@ -13,35 +13,28 @@ const expect = chai.expect
...
-    opts: undefined,
-    init: false
-  }, {
-    name: 'memory',
-    opts: {
-      fs: require('interface-datastore').MemoryDatastore,
-      level: require('memdown'),
-      lock: 'memory'
+  const repos = [
+    {
```
dignifiedquire: please don't change code style, this makes it hard to read the diff

test/node.js
```diff
@@ -13,35 +13,28 @@ const expect = chai.expect
 const IPFSRepo = require('../src')
 
 describe('IPFS Repo Tests on on Node.js', () => {
-  const repos = [{
-    name: 'default',
-    opts: undefined,
-    init: false
```
dignifiedquire: why was this removed?

test/node.js
```diff
@@ -13,35 +13,28 @@ const expect = chai.expect
...
-  }, {
-    name: 'memory',
-    opts: {
-      fs: require('interface-datastore').MemoryDatastore,
-      level: require('memdown'),
-      lock: 'memory'
```
dignifiedquire: why was this removed?

test/node.js
```diff
@@ -13,35 +13,28 @@ const expect = chai.expect
...
     const testRepoPath = path.join(__dirname, 'test-repo')
     const date = Date.now().toString()
     const repoPath = testRepoPath + '-for-' + date
-
     const repo = new IPFSRepo(repoPath, r.opts)
-
     before((done) => {
       series([
-        (cb) => {
-          if (r.init) {
```
dignifiedquire: why are you removing this?

src/index.js
```diff
@@ -34,33 +33,25 @@ class IpfsRepo {
...
     this.closed = true
     this.path = repoPath
-    this.options = Object.assign({
-      sharding: true,
-      lock: 'fs'
-    }, options)
-    this._fsOptions = Object.assign({}, options.fsOptions)
-    const FsStore = this.options.fs
-    this._fsStore = new FsStore(this.path, Object.assign({}, this._fsOptions, {
-      extension: ''
```
dignifiedquire: you can't just remove options that are important

src/index.js
```diff
@@ -34,33 +33,25 @@ class IpfsRepo {
...
-    }))
+    this.options = Object.assign({ lock: 'memory', sharding: true }, options || defaultOptions)
```
dignifiedquire: `lock` still needs to be `fs` not `memory`

src/index.js
diasdavid: I fixed this in #138

src/index.js
justinmchase: Question about this. The `fs` lock appears to not use the datastore but to use fs api's directly. Supposing you were going to use a virtual file storage, and not the local file system, such as s3 would you still want a file system lock? It's not that the file system is completely off limits from writing to but the machine can get taken down and back up and the file system is ephemeral with the instance. Should the lock file also be stored on external storage or is it tied to the life cycle of the process?

Is its purpose to prevent other apps from writing the same files and if so shouldn't the lock actually in the one of the stores?

src/index.js
justinmchase: Yup, that was missed when moving options into default options objects.

test/node.js
justinmchase: For a bad reason. Basically I thought it was an inconsistency in the way the stores were getting created and a bug in this test but now I see that the tests were failing because I dropped the `extension: ''` from the store options which was making the fs based tests fail because it couldn't find the files copied in the before step. Adding the extension back and reverting basically this whole file resolves it.

src/index.js
justinmchase: 👍

justinmchase:
I updated my branch by adding the `extension: ''` option back onto the BlockStore in the constructor which also allowed me to revert the init changes to the tests.

PR #138, opened 2017-06-04
fix: add backwards compatibility and more tests

dignifiedquire:
There were some issues introduced with #136. This should be reverted for now, until it is cleared to not cause any issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ipfs-repo",
-  "version": "0.13.2",
+  "version": "0.13.1",
   "description": "IPFS Repo implementation",
   "main": "src/index.js",
   "browser": {
@@ -40,7 +40,7 @@
   },
   "devDependencies": {
     "aegir": "^11.0.2",
-    "chai": "^4.0.1",
+    "chai": "^3.5.0",
     "dirty-chai": "^1.2.2",
     "lodash": "^4.17.4",
     "memdown": "^1.2.4",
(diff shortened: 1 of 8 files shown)


diasdavid:
Me and @dignifiedquire had a quick chat yesterday, but unfortunately, I wasn't able to follow-up at the time as I was entering a flight.

I've run js-ipfs tests with latest ipfs-repo and they finished successfully, this was a false positive because as @dignifiedquire points out, there were breaking changes with regards to the options object, but since everyone repo uses ipfs-repo defaults, it felt like an internal breaking change.

What we should do in this situation is add backward's compatibility to the old options object, this will enable old custom configs to work if anyone is using any. @dignifiedquire can you refactor your PR to be that? 

With regards to other possible breaking changes. I'm still unsure what those might be as there are no tests that capture those. Mind creating an issue to enumerate those? We should have tests for them in the first place.

diasdavid:
Things to fix:

- [ ] https://github.com/ipfs/js-ipfs-repo/pull/136#discussion_r120016946
- [ ] https://github.com/ipfs/js-ipfs-repo/pull/136#discussion_r120016719
- [ ] https://github.com/ipfs/js-ipfs-repo/pull/136#discussion_r120016840
- [ ] https://github.com/ipfs/js-ipfs-repo/pull/136#discussion_r120016870

We need tests for these also.

package.json
```diff
@@ -40,7 +43,7 @@
   },
   "devDependencies": {
     "aegir": "^11.0.2",
-    "chai": "^4.0.1",
```
diasdavid: Interesting, why the downgrade?

package.json
```diff
@@ -55,15 +58,15 @@
...
-    "interface-datastore": "^0.2.2",
+    "interface-datastore": "^0.2.1",
     "ipfs-block": "~0.6.0",
     "level-js": "timkuijsten/level.js#idbunwrapper",
-    "leveldown": "^1.7.1",
+    "leveldown": "^1.7.0",
     "lock-me": "^1.0.2",
     "multiaddr": "^2.3.0",
-    "safe-buffer": "^5.1.0"
+    "safe-buffer": "^5.0.1"
```
diasdavid: Ah, it was generalized. Mind doing a `ncu -a`?

src/index.js
```diff
@@ -43,21 +44,15 @@ class IpfsRepo {
   constructor (repoPath, options) {
     assert.equal(typeof repoPath, 'string', 'missing repoPath')
 
-    if (options == null) {
-      options = require('./default-options')
```
dignifiedquire: Changing this means that any module bundle will include the default options in the browser always. Not sure this could be an issue or not

pgte: Since webpack and browserify extract dependencies using static analysis, this is always going to get included in the bundle, with or without these lines.

Issue #139, opened 2017-06-28
RFC - Cleansing the 🐉🐉🐉


diasdavid:
It has become extremely complicated to understand what is happening inside this module, the similar names (**store everything), the fact that the API docs were removed in favor of aegir docs but those are not getting published and how things are being initialized. 

The API that we need is only:

```js
const options = {
  path: <path to store the repo in>,
  storageBackends: {
    root: <a interface-datastore compatible class> // default to fs-datastore
    blocks: <a interface-datastore compatible class> // default to fs-datastore
    datastore <a interface-datastore compatible class> // default to level-datastore
  }
}
const repo = new Repo(options)

repo.init(cb) // creates the folder structure
repo.open(cb) // locks the repo 
repo.close(cb) // unlocks the repo

// get and put values at the root of the repo
repo.{get(key), put(key, value}  

// get and put blocks. Knows how to do the sharding and appending .data to the keys
repo.blocks.{get(key), put(key, value}  

// get and put kv. Exposes the full interface-datastore API
repo.datastore.{<full interface-datastore API>}  
```

The `lock` is a special case just to be compatible with go. In Node.js should continue to use fs.

What I'm proposing

- Avoid modules like 'sharding datastore' and just do the sharding from within the `blocks` file here
- Refactor the code, improve comments and flow
- Bring back API docs for the README.
- Write more tests to ensure that there are no breaking changes in the feature that affect how files get stored and how configuration works.

diasdavid:
@kumavis this will require an update on how the 'blockstore' for parity works. But it should be a straightforward update and a simplification of what already exists.

pgte:
@diasdavid what about `repo.config` and `repo.version`?

diasdavid:
from IRC, you made a good point that it would be cumbersome to always be encoding/decoding those values. It makes sense to have those convenient methods. 

There is 3 'convinience' methods in total:
- repo.config
- repo.version
- repo.apiAddr

VictorBjelkholm:
> Bring back API docs for the README.

Feels like this should be a endeavour to either finish the API docs migration we were working on, or revert it as you mentioned. Applies to a lot of modules, not just `js-ipfs-repo` though.

diasdavid:
@VictorBjelkholm there is a whole endeavor around improving our build and release tools -- https://github.com/ipfs/aegir/issues/113 -- which we need to get back to. However, this should not block the ipfs-repo improvements from moving forward.

[...]

PR #140, opened 2017-06-29
API revamp and cleansing

pgte:
Fixes https://github.com/ipfs/js-ipfs-repo/issues/139


README.md
```diff
@@ -137,7 +137,74 @@ This now has created the following structure, either on disk or as an in memory
...
+
+Arguments:
+
+* `path` (string, mandatory): the path for this repo
+* `options` (object, optional): may contain the following values
+  * `lock` (string, defaults to `"fs"` in Node.js, `"memory"` in the browser): what type of lock to use. Lock has to be acquired when opening.
+  * `storageBackends` (object, optional): may contain the following values, which should each be a class implementing the [datastore interface](https://github.com/ipfs/interface-datastore#readme):
+    * `root` (defaults to [`datastore-fs`](https://github.com/ipfs/js-datastore-fs#readme) in Node.js and [`datastore-level`](https://github.com/ipfs/js-datastore-level#readme) in the browser)
+    * `blocks` (defaults to [`datastore-fs`](https://github.com/ipfs/js-datastore-fs#readme) in Node.js and [`datastore-level`](https://github.com/ipfs/js-datastore-level#readme) in the browser)
+    * `datastore` (defaults to `datastore-level`)
```
diasdavid: 👍

README.md
```diff
@@ -137,7 +137,74 @@ This now has created the following structure, either on disk or as an in memory
...
+Creates the necesary folder structure inside the repo.
+
+### repo.open (callback)
+
+Locks the repo.
+
+### repo.close (callback)
+
+Unlocks the repo.
+
```
diasdavid: Let's break the API description into 3 sets:

- Set up - init, open, close and exists
- Repos - Explaining that there are two repos
- Utility/Convinience - apiAddr, config, version..

README.md
```diff
@@ -137,7 +137,74 @@ This now has created the following structure, either on disk or as an in memory
...
+This is contains a full implementation of [the `interface-datastore` API](https://github.com/ipfs/interface-datastore#api).
+
+### repo.config.put(key:string, value, callback)
+
+Set a config value. `value` can be any object that is serializable to JSON.
+
+### repo.config.get(key:string, callback)
+
+Get a config value. `callback` is a function with the signature: `function (err, value)`, wehre the `value` is of the same type that was set before.
```
diasdavid: Currently, the 'replace a key' is being done inside js-ipfs - https://github.com/ipfs/js-ipfs/blob/master/src/core/components/config.js#L10-L55

I like that it is being exposed here, avoids duplication of code if any other module needs to do the same. Let's make sure that once this is done we update js-ipfs to just call the methods directly

src/api-addr.js
```diff
@@ -0,0 +1,44 @@
...
+     */
+    get (callback) {
+      store.get(apiFile, (err, value) => {
+        if (err) {
+          return callback(err)
+        }
+
+        callback(err, value && value.toString())
+      })
+    },
```
diasdavid: since you are doing the check for value, this method can be simplified to:

```
get (callback) {
  store.get(apiFile, (err, value) => callback(err, value && value.toString()))
}
```

src/default-options.js
```diff
@@ -2,8 +2,18 @@
...
+    blocks: require('datastore-fs'),
+    datastore: require('datastore-level')
+  },
+  storageBackendOptions: {
+    root: {
+      extension: ''
+    },
+    blocks: {
+      sharding: true
+    }
```
diasdavid: It's missing the extension `.data`

src/index.js
```diff
@@ -97,60 +75,54 @@ class IpfsRepo {
    */
   open (callback) {
     if (!this.closed) {
-      return callback(new Error('repo is already open'))
+      callback(new Error('repo is already open'))
+      return // early
```
diasdavid: setImmediate the callback

src/index.js
```diff
@@ -97,60 +75,54 @@ class IpfsRepo {
...
-
-        this.blockstore = blockstore(this)
+      (cb) => {
+        log('creating datastore')
+        this.datastore = backends.create('datastore', path.join(this.path, 'datastore'), this.options)
+        log('creating blocks')
+        const blocksBaseStore = backends.create('blocks', path.join(this.path, 'blocks'), this.options)
+        blockstore(
+          blocksBaseStore,
+          this.datastore,
```
diasdavid: Let's reference to `this.datastore` as `this.rootDatastore` (or a name that suggests that it is the root of all mount points, maybe add this as a comment)

src/index.js
```diff
@@ -163,20 +135,23 @@ class IpfsRepo {
...
-    ], (err, res) => {
-      log('init', err, res)
-      if (err) {
-        return callback(err)
-      }
+    parallel(
+      {
+        config: (cb) => this.config.exists(cb),
+        version: (cb) => this.version.check(repoVersion, cb)
+      },
```
diasdavid: Interesting, why making this an object?

src/index.js
pgte: You mean change `this.datastore` to `this.rootDatastore` in the public API, add an alias or simply add a comment?

src/index.js
pgte: Just aesthetics: I prefer to refer to `result.config` rather than `result[0]`.

diasdavid: nice. That I didn't know :)

README.md
pgte: Done.

src/index.js
diasdavid: ah, wait.

this.datastore is /datastore? If so, why does /blocks need it?
if this.datastore is just /, then it makes sense to call it rootDatastore

pgte: Ah, it's legacy, it used to need it but now doesn't. Good catch, removing it.

pgte: removed

README.md
pgte: Yes, it's in https://github.com/ipfs/js-ipfs/pull/887/files#diff-10e3e7645db7852759507e460173d205

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+#### repo.close (callback)
+
+Unlocks the repo.
+
+#### repo.exists (callback)
+
+Tells whether this repo exists or not. Callsback with `(err, bool)`.
+
+### Repos
+
```
diasdavid: let's add a line here saying "root repo"

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+* `key` can be a buffer, a string or a [Key](https://github.com/ipfs/interface-datastore#keys).
+
+#### repo.get (key, callback)
+
+Get a value at the root of the repo.
+
+* `key` can be a buffer, a string or a [Key](https://github.com/ipfs/interface-datastore#keys).
+* `callback` is a callback function `function (err, result:Buffer)`
+
```
diasdavid: let's add a line here saying "blocks/blockstore" and link it to the ipfs-block class

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+* `cid` is the content id of [type CID](https://github.com/ipld/js-cid#readme).
+* `callback` is a callback function `function (err, result:Buffer)`
+
+#### repo.datastore
+
+This is contains a full implementation of [the `interface-datastore` API](https://github.com/ipfs/interface-datastore#api).
+
+
+### Convenience
```
diasdavid: Maybe call this `utils`? They are convenience methods but typically these are listed as utils

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+
+#### repo.config
+
```
diasdavid: Explain why using repo.config is better than repo.set('config')

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+```js
+repo.config.set('a.b.c', 'c value', (err) => {
+  if (err) throw err
+  repo.config.get((err, config) => {
+    if (err) throw err
+    assert.equal(config.a.b.c, 'c value')
+  })
+})
+``
```
diasdavid: missing `

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+
+Sets the repo version
+
+#### repo.apiAddr
+
+#### repo.apiAddr.get (callback)
+
+Gets the API address.
+
+#### repo.apiAddr.set (value:string, callback)
```
diasdavid: Should take a String or multiaddr, but should always be a valid multiaddr

diasdavid:
@justinmchase I believe you are going to like the results of this PR as it will make way more obvious and testable that S3 backend you are looking for :)

README.md
```diff
@@ -137,7 +137,142 @@ This now has created the following structure, either on disk or as an in memory
...
+
+##### repo.config.get(key:string, callback)
+
+Get a config value. `callback` is a function with the signature: `function (err, value)`, wehre the `
+value` is of the same type that was set before.
+
+* `key` is a string specifying the object path. Example:
+
+```js
+repo.config.set('a.b.c', (err, value) => {
```
VictorBjelkholm: Should be `.get`

src/index.js
```diff
@@ -221,33 +195,42 @@ class IpfsRepo {
...
-   * @returns {void}
-   */
-  apiAddress (callback) {
-    this._fsStore.get(apiFile, (err, rawAddr) => {
-      if (err) {
-        return callback(err)
-      }
+function ignoringAlreadyOpened (cb) {
+  return (_err) => {
+    let err = _err
```
VictorBjelkholm: Not 100% I understand this type of assignment, why not name it directly `err` and just use that?

pgte: The only reason is that I think that assigning to an argument is a bad practise..

README.md
pgte: done

README.md
pgte: done

README.md
```diff
@@ -137,7 +137,152 @@ This now has created the following structure, either on disk or as an in memory
...
+#### repo.apiAddr.get (callback)
+
+Gets the API address.
+
+#### repo.apiAddr.set (value, callback)
+
+Sets the API address.
+
+* `value` should be a [Multiaddr](https://github.com/multiformats/js-multiaddr) or a String representing a valid one.
+
```
dignifiedquire: seems to be missing `exists`

src/api-addr.js
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Set the current configuration for this repo.
+     *
+     * @param {Object} value - the api address to be written
+     * @param {function(Error)} callback
+     * @returns {void}
+     */
+    set (value, callback) {
+      store.put(apiFile, Buffer.from(value.toString()), callback)
```
dignifiedquire: `Buffer.from` is not available in node@4 you will need to include `safe-buffer`

src/config.js
```diff
@@ -1,44 +1,69 @@
...
         let config
         try {
-          config = JSON.parse(value.toString())
+          config = JSON.parse(encodedValue.toString())
         } catch (err) {
           return callback(err)
         }
-        callback(null, config)
+        if (key !== undefined && !_has(config, key)) {
+          callback(new Error('Key does not exist in config'))
```
dignifiedquire: would be nice to say which key doesn't exist, i.e.`Key "foo" does not exist in config`

src/index.js
```diff
@@ -221,33 +195,42 @@ class IpfsRepo {
...
 
-      callback(null, new Multiaddr(rawAddr.toString()))
-    })
+function ignoringNotFound (cb) {
+  return (_err) => {
+    let err = _err
+    if (err && err.message.startsWith('ENOENT')) {
+      err = null
+    }
+    cb(err)
```
dignifiedquire: These to ignore calls could share some code, sth like this

```js
const ignoreIf = (cond) => (cb) => (err) => {
  cb(err && !cond(err) ? err : null)
}

const ingoreNotFound = ignoreIf((err) => err.message.startsWith('ENOENT'))
const ignoreAlreadyOpen = ignoreIf((err) => err.message === 'Already open'))
```

src/blockstore.js
```diff
@@ -31,8 +30,27 @@ const cidToDsKey = (cid) => {
...
 }
 
-module.exports = (repo) => {
-  const store = new NamespaceStore(repo.store, blockPrefix)
+module.exports = (filestore, options, callback) => {
+  maybeWithSharding(filestore, options, (err, store) => {
+    if (err) {
+      callback(err)
+      return // early
+    }
```
justinmchase: This section can be a single line, which standard will allow:
```
if (err) return callback(err)
```

src/default-options.js
```diff
@@ -2,8 +2,19 @@
...
+  },
+  storageBackendOptions: {
+    root: {
+      extension: ''
+    },
+    blocks: {
+      sharding: true,
+      extension: '.data'
+    }
+  }
```
justinmchase: 👍

README.md
justinmchase: It would be nice to have a brief description of what the role of each of these datastores are at a high level.

src/index.js
```diff
@@ -1,30 +1,27 @@
...
 const defaultOptions = require('./default-options')
 
 const log = debug('repo')
 
-const apiFile = new Key('api')
-const flatfsDirectory = 'blocks'
-const levelDirectory = 'datastore'
+const lockers = {
+  memory: require('./lock-memory'),
+  fs: require('./lock')
```
justinmchase: Shouldn't the locks also be a datastore? What if you are running on an ephemeral server and it goes down and back up, would the locks need to be preserved or is it ok that they are lost when the server is replaced? For example if your server was a Docker container and you update the instance.

src/index.js
justinmchase: A couple things here, its fairly standard to have a code property on an error in addition to message for this sort of situation (e.g. [fs.stats documentation](https://nodejs.org/dist/latest-v7.x/docs/api/fs.html#fs_fs_stat_path_callback)).

Also, the parameter should be called error for convention purposes, standard will do some enforcement around variables of that name specifically as well.

I would probably write this as:
```js
function ignoringAlreadyOpened (cb) {
  return (err) => err && err.code === 'AOPEN' ? callback() : callback(err)
}
```

This would imply you are adding the code `'AOPEN'` too of course. You don't want to really assume the error message is going to be English.

src/index.js
justinmchase: Same with this, you should probably have error codes on these.

README.md
pgte: done

README.md
pgte: done

README.md
pgte: good catch, fixed

README.md
pgte: done

src/api-addr.js
pgte: done

src/index.js
dignifiedquire: Lock can't be a simple datastore as it uses os specific system calls on the filesystem. Though the memory lock could use something else as backend if needed. It might be good to allow passing in ones own lock module if one wants to instead of using one of the specified

src/blockstore.js
dignifiedquire: but our linting rules do not, right @diasdavid ;)

src/index.js
dignifiedquire: We currently don't have unified error handling on the datastore interfaces. These errors come directly from them which is why have to make string matching right now. I agree it would be better to have subclasses of `Error` + error codes for checking these but these need to be implemented first in all the datastores.

src/api-addr.js
pgte: fixed

src/blockstore.js
diasdavid: ❤️ {}

diasdavid: I'm not against to `if (err) { return callback(err) }` if the goal is to save a couple of lines.

pgte:
I think I've addressed every concern here. Please tell me if there's anything else I need to do to make this happen, as I depend on this. :)

README.md
diasdavid: - [ ] @pgte

README.md
```diff
@@ -137,7 +137,152 @@ This now has created the following structure, either on disk or as an in memory
...
+Locks the repo.
+
+#### repo.close (callback)
+
+Unlocks the repo.
+
+#### repo.exists (callback)
+
+Tells whether this repo exists or not. Callsback with `(err, bool)`.
```
diasdavid: Missing space

README.md
```diff
@@ -137,7 +137,152 @@ This now has created the following structure, either on disk or as an in memory
...
+
+### Repos
+
+Root repo:
+
+#### repo.put (key, value:Buffer, callback)
```
diasdavid: Unnecessary spaces.

Also, can we put these between ``, helps readability on the readme.

README.md
```diff
@@ -137,7 +137,152 @@ This now has created the following structure, either on disk or as an in memory
...
+
+##### repo.config.set(key:string, value, callback)
+
+Set a config value. `value` can be any object that is serializable to JSON.
+
+* `key` is a string specifying the object path. Example:
+
+```js
+repo.config.set('a.b.c', 'c value', (err) => {
+  if (err) throw err
```
diasdavid: please use `{}`, even if they are all in the same line so that it follows our agreed codestyle

README.md
```diff
@@ -137,7 +137,152 @@ This now has created the following structure, either on disk or as an in memory
...
+  repo.config.get((err, config) => {
+    if (err) throw err
+    assert.equal(config.a.b.c, 'c value')
+  })
+})
+```
+
```
diasdavid: extra `\n`

README.md
diasdavid: - [ ] @pgte

README.md
pgte: done

README.md
pgte: fixed

pgte:
@diasdavid listed dependents in awesome endeavour PR: https://github.com/ipfs/js-ipfs/pull/887

pgte:
@diasdavid https://github.com/ipfs/js-ipfs-repo/pull/140/commits/85a281073376ac5cafbc3e3f20a1d3e7cb739c02 and https://github.com/ipfs/js-ipfs-repo/pull/140/commits/c5695edfecf210d2b38fb9702b991b386f660285 test this against a repo generated by go-ipfs v0.4.10.

PR #148, opened 2017-11-08
Fix error with key generation on Windows

GhostfromTexas:
Fix by Balanced Media Technology

The key being generated is used as a file path, and was only supportive of a unix-based pathing system (forward vs backwards slashes)

Changed to use Node Path's path.sep to provide OS specific slashes for pathing.

### DIFF
--- a/src/blockstore.js
+++ b/src/blockstore.js
@@ -8,6 +8,7 @@ const Block = require('ipfs-block')
 const setImmediate = require('async/setImmediate')
 const reject = require('async/reject')
 const CID = require('cids')
+const path = require('path')
 
 /**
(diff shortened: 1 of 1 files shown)


richardschneider:
Thanks for the PR.  This issue is much deeper in the stack and is being addressed by https://github.com/ipfs/js-ipfs/issues/1017.  Basically a Key path should **NOT** be os-specific and always use the forward slash.

Hopefully all will be fixed in a few days

GhostfromTexas:
Understood! Thanks for the comment. Right now it seems to be working for us and atm we're all windows based.

Will definitely keep an eye out for the real fix soon :)

PR #159, opened 2018-01-30
Implement .stat() function

hacdias:
To implement this, I take a look at [Go's repo.stat implementation](https://github.com/ipfs/go-ipfs/blob/4c392928b48db4cdd8853234202ab97c78e3d56b/core/corerepo/stat.go) and I basically do the same on each one except on `repoSize`.

- For `numObjects` I use the number of blocks on the Blockstore
- For `repoPath` I just used `this.path`
- For `version` I use `this.version.get`
- For `storageMax` I use the config `Datastore.StorageMax` or the maximum integer number.
- For `repoSize` I sum the number of bytes of the blocks and their keys from the Blockstore + the same for the Datastore and the Keystore. This doesn't count with `api` and ´config` files' size but it won't differ much from the real size. 

I also added a simple test.

### DIFF
--- a/src/blockstore.js
+++ b/src/blockstore.js
@@ -8,6 +8,7 @@ const Block = require('ipfs-block')
 const setImmediate = require('async/setImmediate')
 const reject = require('async/reject')
 const CID = require('cids')
+const pull = require('pull-stream')
 
 /**
  * Transform a raw buffer to a base32 encoded key.
@@ -49,6 +50,19 @@ function maybeWithSharding (filestore, options, callback) {
 
 function createBaseStore (store) {
   return {
+    /**
+     * Query the store.
+     *
+     * @param {object} query
+     * @param {function(Error, Array)} callback
+     * @return {void}
+     */
+    query (query, callback) {
+      pull(
+        store.query(query),
+        pull.collect(callback)
+      )
+    },
     /**
      * Get a single block by CID.
      *
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,8 @@ const each = require('async/each')
 const assert = require('assert')
 const path = require('path')
 const debug = require('debug')
+const Big = require('big.js')
+const pull = require('pull-stream')
 
 const backends = require('./backends')
 const version = require('./version')
@@ -16,6 +18,8 @@ const blockstore = require('./blockstore')
 const defaultOptions = require('./default-options')
 
 const log = debug('repo')
+
+const noLimit = Number.MAX_SAFE_INTEGER
 
 const lockers = {
   memory: require('./lock-memory'),
@@ -200,6 +204,81 @@ class IpfsRepo {
   exists (callback) {
     this.version.exists(callback)
   }
+
+  /**
+   * Get repo status.
+   *
+   * @param {Object}  options
+   * @param {Boolean} options.human
+   * @param {function(Error, Object)} callback
+   * @return {void}
+   */
+  stat (options, callback) {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+
+    options = Object.assign({}, {human: false}, options)
+
+    parallel({
+      storageMax: (cb) => this.config.get('Datastore.StorageMax', (err, max) => {
+        if (err) {
+          cb(null, new Big(noLimit))
+        } else {
+          cb(null, new Big(max))
+        }
+      }),
+      version: (cb) => this.version.get(cb),
+      blocks: (cb) => this.blocks.query({}, (err, list) => {
+        list = list || []
+
+        const count = new Big(list.length)
+        let size = new Big(0)
+
+        list.forEach(block => {
+          size = size
+            .plus(block.value.byteLength)
+            .plus(block.key._buf.byteLength)
+        })
+
+        cb(err, {
+          count: count,
+          size: size
+        })
+      }),
+      datastore: (cb) => getSize(this.datastore, cb),
+      keys: (cb) => getSize(this.keys, cb)
+    }, (err, results) => {
+      if (err) return callback(err)
+
+      let size = results.blocks.size
+        .plus(results.datastore)
+        .plus(results.keys)
+
+      if (options.human) {
+        size = size.div(1048576)
+      }
+
+      callback(null, {
+        repoPath: this.path,
+        storageMax: results.storageMax,
+        version: results.version,
+        numObjects: results.blocks.count,
+        repoSize: size
+      })
+    })
+  }
+}
+
+function getSize (queryFn, callback) {
+  pull(
+    queryFn.query({}),
+    pull.reduce((sum, block) => {
+      return sum
+        .plus(block.value.byteLength)
+        .plus(block.key._buf.byteLength)
+    }, new Big(0), callback))
 }
 
 module.exports = IpfsRepo
--- a/README.md
+++ b/README.md
@@ -282,6 +282,20 @@ Sets the API address.
 
 * `value` should be a [Multiaddr](https://github.com/multiformats/js-multiaddr) or a String representing a valid one.
 
+### `repo.stat ([options], callback)`
+
+Gets the repo status. 
+
+`options` is an object which might contain the key `human`, which is a boolean indicating whether or not the `repoSize` should be displayed in MiB or not.
+
+`callback` is a function with the signature `function (err, stats)`, where `stats` is an Object with the following keys:
+
+- `numObjects`
+- `repoPath`
+- `repoSize`
+- `version`
+- `storageMax`
+
 ## Notes
 
 - [Explanation of how repo is structured](https://github.com/ipfs/js-ipfs-repo/pull/111#issuecomment-279948247)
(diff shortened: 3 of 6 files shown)


src/blockstore.js
```diff
@@ -49,6 +50,21 @@ function maybeWithSharding (filestore, options, callback) {
...
+     * Query the store.
+     *
+     * @param {object} query
+     * @param {function(Error, Array)} callback
+     * @return {void}
+     */
+    query (query, callback) {
+      pull(
+        store.query(query),
+        pull.collect((err, list) => {
```
pgte: I think you can pass callback directly to `pull.collect` here.

hacdias:
@pgte @dignifiedquire @diasdavid I think this is done and the tests are passing 😄 Could you take a look and if it looks good, merge and publish a new version?

Needed for https://github.com/ipfs/js-ipfs/issues/1197

src/index.js
```diff
@@ -200,6 +204,72 @@ class IpfsRepo {
...
   }
+
+  /**
+   * Get repo status.
+   *
+   * @param {function(Error, Object)} callback
+   * @return {void}
+   */
+  stat (callback) {
+    waterfall([
```
dignifiedquire: why not do the stats in parallel, and aggregate them to an object at the end?

hacdias: Good idea!

hacdias: Although there are some that use the same variable. Wouldn't that cause some problem eventually?

pgte: Here you're using waterfall to pass the accumulator, but since, by what I gather, each step is creating it's own entry, perhaps you could use [`async/parallel`](https://caolan.github.io/async/docs.html#parallel) with the object arguments, collecting them all automatically in the final callback.

src/index.js
```diff
@@ -200,6 +204,72 @@ class IpfsRepo {
...
+
+        cb(null, stats)
+      })
+    ], callback)
+  }
+}
+
+function getSize (queryFn, callback) {
+  pull(
+    queryFn.query({}),
```
pgte: Try to avoid buffering: I think you can reduce this iteratively instead of buffering the entire list.

pgte: My comment refers to line below where you do `pull.collect`.

src/index.js
pgte: If you require some of them to be series, you can still make the all series, but perhaps use `async/series` with the object argument to make this less explicit.

hacdias: Multiple steps use the `repoSize`. If it was parallel I think I could end up messed up.

src/blockstore.js
hacdias: Nice catch! Thanks 😄

src/index.js
pgte: So perhaps you could calculate that first, and then proceed to make the rest parallel?

src/index.js
hacdias: Sorry, but I didn't get this. How can I do this iteratively?

src/index.js
hacdias: Yeah!

src/index.js
pgte: Since the query is a pull-stream, you can use `pull.reduce` like this:

```js
pull(
  source,
  pull.reduce((val) => compute(val), 0, callback))
```

hacdias: Oh, didn't know that. Thanks

hacdias:
@dignifiedquire @pgte changes made!

src/index.js
```diff
@@ -200,6 +204,87 @@ class IpfsRepo {
...
+  stat (options, callback) {
+    if (typeof options === 'function') {
+      callback = options
+      options = {}
+    }
+
+    options = Object.assign({}, {human: false}, options)
+
+    parallel([
+      (cb) => cb(null, this.path),
```
dignifiedquire: no need for this, just use it directly at the end

src/index.js
```diff
@@ -200,6 +204,87 @@ class IpfsRepo {
...
+        if (err) {
+          cb(null, new Big(noLimit))
+        } else {
+          cb(null, new Big(max))
+        }
+      }),
+      (cb) => this.version.get(cb),
+      (cb) => waterfall([
+        (cb) => cb(null, {
+          repoSize: new Big(0)
```
dignifiedquire: no need, just use the value directly

src/index.js
```diff
@@ -200,6 +204,87 @@ class IpfsRepo {
...
+        }),
+        (stats, cb) => getSize(this.datastore, (err, size) => {
+          stats.repoSize = stats.repoSize.plus(size)
+          cb(err, stats)
+        }),
+        (stats, cb) => getSize(this.keys, (err, size) => {
+          stats.repoSize = stats.repoSize.plus(size)
+          cb(err, stats)
+        }),
+        (stats, cb) => {
```
dignifiedquire: the below doesn't need to be a callback

src/index.js
```diff
@@ -200,6 +204,87 @@ class IpfsRepo {
...
+        } else {
+          cb(null, new Big(max))
+        }
+      }),
+      (cb) => this.version.get(cb),
+      (cb) => waterfall([
+        (cb) => cb(null, {
+          repoSize: new Big(0)
+        }),
+        (stats, cb) => this.blocks.query({}, (err, list) => {
```
dignifiedquire: the size fetching below can all happen in parallel, just need to sum at the end

hacdias:
@dignifiedquire @pgte hopefully everything is nicer now 😃

test/repo-test.js
```diff
@@ -99,4 +99,32 @@ module.exports = (repo) => {
...
+    it('get stats', (done) => {
+      repo.stat((err, stats) => {
+        expect(err).to.not.exist()
+        expect(stats).to.exist()
+        expect(stats).to.have.property('numObjects')
+        expect(stats).to.have.property('version')
+        expect(stats).to.have.property('repoPath')
+        expect(stats).to.have.property('repoSize')
+        expect(stats).to.have.property('storageMax')
+        done()
```
dignifiedquire: just realized it now, these don’t test that they are the right numbers, so you should at least check thatj they are greater 0 or similar

src/index.js
hacdias: Done! 😄

src/index.js
```diff
@@ -200,6 +204,85 @@ class IpfsRepo {
...
+            .plus(block.key._buf.byteLength)
+        })
+
+        cb(err, {
+          count: count,
+          size: size
+        })
+      }),
+      (cb) => getSize(this.datastore, (err, size) => {
+        cb(err, new Big(size))
```
dignifiedquire: this can be simplified as getSize already returns a bigint

src/index.js
hacdias: Done!

test/repo-test.js
hacdias: @dignifiedquire done! had to move the tests to another file so they run after the other tests add some data to the repo 😄

src/index.js
hacdias: Done!

src/index.js
```diff
@@ -200,6 +204,81 @@ class IpfsRepo {
...
+      (cb) => getSize(this.keys, cb)
+    ], (err, results) => {
+      if (err) return callback(err)
+
+      let size = results[2].size
+        .plus(results[3])
+        .plus(results[4])
+
+      if (options.human) {
+        size = size.div(1048576)
```
dignifiedquire: why not use something like https://www.npmjs.com/package/pretty-bytes instead?

hacdias: The spec says MiB so it must be in base 2 and not 10. Unless we could find another package. But why importing another package if we can just divide by that number?

hacdias:
Cool! Thanks @dignifiedquire 😄

src/index.js
dignifiedquire: Ah okay, the package only makes sense if the output should be dynamically adjusted depending on the size.

src/index.js
```diff
@@ -200,6 +204,81 @@ class IpfsRepo {
...
+        })
+
+        cb(err, {
+          count: count,
+          size: size
+        })
+      }),
+      (cb) => getSize(this.datastore, cb),
+      (cb) => getSize(this.keys, cb)
+    ], (err, results) => {
```
pgte: Perhaps it would be easier to use the object notation of async/parallel, something like this:

```js
parallel({
  a: (cb) => cb(null, 1),
  b: (cb) => cb(null, 2),
  c: (cb) => cb(null, 3)
}, (err, results) => {
  if  (err) { handleError(err) }
  else {
    console.log(results) // { a: 1, b: 2, c: 3}
  }
})
```

pgte:
@hacdias except for the async/parallel remark, LGTM!

src/index.js
hacdias: Yeah, it would probably get a bit more clear. I'll do that.

hacdias:
Done and the tests are passing. Could any of you merge this and release a new version when possible? 😃

diasdavid:
`v0.18.7` released :)

hacdias:
Cool! Thanks

Issue #161, opened 2018-04-06
Allow the repo locker to be passed by configuration


jacobheun:
Currently the only available lockers are fs and memory. It's currently not possible to set a locker via configuration. Providing support for this would allow new datastores to provide their own locking mechanism.

Perhaps updating the Repo initialization to include a `locker` property would be a good approach:
```
const repo = new Repo({
  locker: require('./my-locker')
})
```
Then when setting `this._locker` internally, it could check for either  this.options.locker, or the `fs|memory` keys in this.options.lock

connects to ipfs/js-datastore-s3#1

diasdavid:
Thanks for bringing this up. I agree there is a good case to make the lock file optional and even avoiding using fs-ext to lock on fs (which causes problems to electron). From my side:

- 👍 on making the lock optional and configurable.
- 👍 on abandoning fs-ext.

PR #162, opened 2018-04-09
feat(lock): allow for custom lock options

jacobheun:
* Allows users to create custom repo locks, such as to support the new s3 datastore
* ~~Allows users to have no repo locking~~

resolves ipfs/js-ipfs-repo#161

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -44,8 +44,7 @@ class IpfsRepo {
     this.closed = true
     this.path = repoPath
 
-    this._locker = lockers[this.options.lock]
-    assert(this._locker, 'Unknown lock type: ' + this.options.lock)
+    this._locker = this._getLocker()
 
     this.root = backends.create('root', this.path, this.options)
     this.version = version(this.root)
@@ -88,7 +87,7 @@ class IpfsRepo {
     waterfall([
       (cb) => this.root.open(ignoringAlreadyOpened(cb)),
       (cb) => this._isInitialized(cb),
-      (cb) => this._locker.lock(this.path, cb),
+      (cb) => this._openLock(this.path, cb),
       (lck, cb) => {
         log('aquired repo.lock')
         this.lockfile = lck
@@ -121,7 +120,7 @@ class IpfsRepo {
       }
     ], (err) => {
       if (err && this.lockfile) {
-        this.lockfile.close((err2) => {
+        this._closeLock((err2) => {
           if (!err2) {
             this.lockfile = null
           } else {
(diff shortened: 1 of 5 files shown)


diasdavid:
@pgte @dignifiedquire can I get a review from you on this feature?

test/options-test.js
```diff
@@ -28,6 +28,26 @@ describe('custom options tests', () => {
     const repo = new Repo(repoPath)
     expect(repo.options).to.deep.equal(expectedRepoOptions())
   })
+
+  it('allows for no locker', () => {
+    const repo = new Repo(repoPath, {
+      locker: false
```
dignifiedquire: should also test for `undefined` options

test/node.js
```diff
@@ -25,6 +41,16 @@ describe('IPFS Repo Tests onNode.js', () => {
       lock: 'memory'
     },
     init: true
+  }, {
+    name: 'custom locker',
```
dignifiedquire: There is no proper locking tests currently, so it would be good to test various locking scenarios in addition to the regular tests

src/index.js
```diff
@@ -135,6 +134,65 @@ class IpfsRepo {
...
   }
 
+  /**
+   * Returns the repo locker to be used. Null will be returned if no locker is requested
+   *
+   * @private
+   * @returns {Locker}
+   */
+  _getLocker () {
+    if (this.options.locker === false) {
```
dignifiedquire: This way of options is :-1: from my side. This forces `locker` to be of multiple types, so disabling the locker should check for `!= null`  and be if it is not set, rather than a boolean

src/index.js
```diff
@@ -135,6 +134,65 @@ class IpfsRepo {
...
+  /**
+   * Creates a lock on the repo if a locker is specified. The lockfile object will
+   * be returned in the callback if one has been created.
+   *
+   * @param {string} path
+   * @param {function(Error, lockfile)} callback
+   * @returns {void}
+   */
+  _openLock (path, callback) {
+    if (this._locker) {
```
dignifiedquire: also 👎 on being able to disable the locking, this is just asking for trouble

jacobheun: I am fine with leaving locking required, the custom locker is the needed piece. It's fairly easy to subvert locking at the moment even if we don't explicitly allow it to be disabled. With custom datastore support there is a potential use case for users wanting to handle locking at the object level, rather than the repo level.

If we decide to leave it required, I can update the documentation when I add in the custom locker docs to encourage users to use locking appropriate for their use case and custom datastores, such as storing their lock on S3 if their datastore is there.

src/index.js
Mr0grog: I know multiple types for an option isn’t especially desirable, but it feels to me like just using `lock` for both custom and built-in types instead of adding a new option would be *much* nicer from the perspective of a user. Maybe we could move towards the type being a Locker object instead of a string? You could support built-ins with:

```js
// Expose the built-in collection as a frozen static var on Repo
new Repo({ lock: Repo.lockers.memory })
// OR
new Repo({ lock: require('ipfs-repo/lock-memory') })
```

…though we’d probably want to keep string support as a deprecated feature for a little while.

(Found my way here after reviewing ipfs/js-ipfs#1303; I hope I’m not butting in.)

jacobheun:
@dignifiedquire @diasdavid

I've made a few updates after everyones feedback:
1) Documentation has been added to the Readme to help users see how to add a custom lock
2) I have removed the ability to disable locking, so this feature update can focus on the customization aspect only. This should give users the ability to add appropriate locking based on their datastore types.
3) I've added a suite to test that locks are being applied and that locking enforces one Node per Repo. 

I also updated the repo options to allow for overloading of the `lock` option as a string or Lock, instead of a separate property. As @Mr0grog mentioned, I also think this is more friendly for users. I noted that specifying `lock` as a string is *Deprecated*, but it is still currently supported to allow for backwards compatibility.

jacobheun:
@Mr0grog I updated things based on your comments, and I agree we should check for the `lock.close` method. I updated the code and added a test to verify it. If the assertion error is triggered, the internal lockfile reference won't be set, so `_closeLock` won't get called.

PR #172, opened 2018-07-19
Support for swarm keys

jacobheun:
This gives ipfs-repo the ability to return swarm.key files that are stored in its root. This is necessary for autostarting an ipfs node with a private network swarm.

This will make it possible to implement the [pnet protector](https://github.com/libp2p/js-libp2p-pnet) in js-ipfs.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ const path = require('path')
 const debug = require('debug')
 const Big = require('bignumber.js')
 const pull = require('pull-stream')
+const Key = require('interface-datastore').Key
 
 const backends = require('./backends')
 const version = require('./version')
@@ -321,6 +322,22 @@ class IpfsRepo {
         numObjects: results.blocks.count,
         repoSize: size
       })
+    })
+  }
+
+  /**
+   * Gets the swarm.key buffer, if it exists
+   *
+   * @param {function(Error, Buffer)} callback
+   * @returns {void}
+   */
+  swarmKey (callback) {
+    const swarmKeyKey = new Key('swarm.key')
+    this.root.get(swarmKeyKey, (err, swarmKeyBuffer) => {
+      if (err) {
+        return callback(new Error(ERRORS.SWARM_KEY_NOT_FOUND))
+      }
+      callback(null, swarmKeyBuffer)
     })
   }
 }
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 * text=auto
 test/test-repo/** text eol=lf
+test/fixtures/** text eol=lf
--- a/src/errors/index.js
+++ b/src/errors/index.js
@@ -1,3 +1,4 @@
 'use strict'
 
 exports.ERR_REPO_NOT_INITIALIZED = 'ERR_REPO_NOT_INITIALIZED'
+exports.SWARM_KEY_NOT_FOUND = 'No swarm key found'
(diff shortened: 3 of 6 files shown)


jacobheun:
@VictorBjelkholm can you disable appveyor for js-ipfs-repo? I've remove the non-jenkins files.

jacobheun:
@VictorBjelkholm can we get Appveyor shut off for js-ipfs-repo?

src/index.js
```diff
@@ -331,6 +332,22 @@ class IpfsRepo {
...
+
+  /**
+   * Gets the swarm.key buffer, if it exists
+   *
+   * @param {function(Error, Buffer)} callback
+   * @returns {void}
+   */
+  swarmKey (callback) {
+    const swarmKeyKey = new Key('swarm.key')
+    this.root.get(swarmKeyKey, (err, swarmKeyBuffer) => {
```
richardschneider: All keys should be on the [keychain](https://github.com/libp2p/js-libp2p-keychain)

jacobheun: @richardschneider this implementation is based off of the current implementation for private network keys in go, https://github.com/ipfs/go-ipfs/blob/07feeec9c4f40f59efb57d7e16739a35b4c42d5b/repo/fsrepo/fsrepo.go#L681-L695, which are not store in the key store.

Looking at the keychain it doesn't look like it currently has the ability to store this key type. Here is the working private network spec for reference https://github.com/libp2p/specs/blob/b1365bedcd46442074fbf96610f66b0663be088a/pnet/Private-Networks-PSK-V1.md

daviddias:
@jacobheun what's missing to land this?

jacobheun:
@daviddias we need to confirm the path forward, given the concern that this is not leveraging the libp2p Keystore. There are two primary options as I see it:

**Behave like go-ipfs**
go-ipfs currently checks the repo for the existence of `swarm.key` in its root. If it exists, it loads the key and privatizes all connections. The advantage of doing this is that we behave the same way as go-ipfs. The disadvantage is that we're not using the Keystore. If we do this we could do one of two things:
1. Rebase and merge this PR with the latest async changes, then update js-ipfs
2. Close this PR and keep the logic isolated to js-ipfs. This would allow us to more easily migrate to Keystore later if we wanted to, without leaving any PSK code in js-ipfs-repo.

**Use the Keystore**
The main drawback here is that this will not behave like go-ipfs. Rather than adding your key to the root of your IPFS repo, you would add it to the Keystore, such as `jsipfs key import -t psk -f ./swarm.key SwarmKey`. Libp2p could then check the Keystore for the existence of the swarm key. In order to do this we would need to:
1. Update libp2p-keystore to support importing and storing PSKs
2. Update the js-ipfs cli to support importing PSKs
3. Update js-ipfs to configure libp2p with the PSK from the Keystore


IMO, we should match go-ipfs and do option 2.  Additionally, we should open an issue to discuss migrating both implementations to using the Keystore in the future, as that should be the approach we use long term.

Just a note, it is possible to privatize js-ipfs today, as libp2p can programmatically be customized, however, it can't be done using just the CLI client.

cc: @alanshaw

alanshaw:
> IMO, we should match go-ipfs and do option 2. Additionally, we should open an issue to discuss migrating both implementations to using the Keystore in the future, as that should be the approach we use long term.

+1

achingbrain:
@jacobheun do we still need this PR?

jacobheun:
No, we have the ability to create pnets by customizing libp2p in js-ipfs. If we want to support that being built in by just adding a swarm.key file, we can modify js-ipfs to check for the existence of the key on boot. Closing this.

PR #173, opened 2018-07-21
WIP: fix: interoperability with go repo

### LINKED ISSUE #163: Error: required Datastore.Spec entry missing form config file
Hi there :)

### TL;DR
`js-ipfs-repo` seems to miss some entries in the `config` file.

### What
I'm working with `orbit-db-cli`  and trying to examine the `ipfs repo` with `go-ipfs` results in this error message. `js-ipfs` doesn't care about it and works fine.

I'd liked to be able to examine the `repo` with `go-ipfs` (since it has a bit more features ), but adding the default `config` file to it, does not solve the issue.

Any ideas what I can do? Adding the entry to the config file should fix the issue, but I might made some stupid mistake I'm not aware of.

For more background see [this issue](https://github.com/orbitdb/orbit-db-cli/issues/16).

vasco-santos:
When a `go-ipfs` daemon runs the `init` command, it creates a `datastore_spec` file and the `config` file has the `datastore` property on it. This way, when a go daemon stars, it expects to have those two in its repo.

Aiming to have interoperability between go and ipfs repo ([interop#26](https://github.com/ipfs/interop/pull/26) and [interop#28](https://github.com/ipfs/interop/pull/28)), the `js-ipfs` repo has to have both in its repo.

- [ ] Add datastore default config to `js-ipfs` [js-ipfs#1460](https://github.com/ipfs/js-ipfs/issues/1460) (instead of directly here)
- [ ] Add running test in `ipfs/interop` for the repo interoperability [interop#29](https://github.com/ipfs/interop/pull/29)

Fixes #163


vasco-santos:
Yeah @jacobheun ! That is one of the problems that I would like to discuss. The ability to customize may be problematic because `go-ipfs` has different types of datastores, which are still not available on `js-ipfs`.

jacobheun:
That's the downside of customizing, you start losing compatibility. Out of the box we want js and go to interop without issue, but as people change that default configuration we can't guarantee that's the case anymore. I think it's something we can put in the readme when we update it for this to say, "If you change this, you risk losing ipfs-repo interoperability with other languages (go, etc)".

If we don't recognize the spec or we can't find the matching datastores, sharding methods, etc, it should just throw an incompatibility error and exit. No touching the data if it doesn't know how to handle it properly.

vasco-santos:
After a productive call with @jacobheun we created an [endeavour issue](https://github.com/ipfs/js-ipfs/issues/1467) to tackle all the tasks regarding the repo interoperability. This issue aims to tackle the following set of tasks:

#### 3) [ipfs/js-ipfs-repo](https://github.com/ipfs/js-ipfs-repo)

- [ ] **1.** Throw an error on invalid / inconsistent configs
  - **INIT**: create the repo based on the config received (verify if the storageBackends exists and if so, guarantee that it matches the spec). If is is not possible to create the repo an error should be thrown to inform that the options received are not valid.
  - **Start**: verify if the repo that exists is compliant with the configuration. If not, throw an error
- [ ] **2.**  transition for using spec instead of the current programatic approach (support both to avoid breaking existing projects, but deprecate the current format)

PR #177, opened 2018-10-03
improved error message

steefmin:
Fixes https://github.com/ipfs/js-ipfs/issues/1411

### DIFF
--- a/src/version.js
+++ b/src/version.js
@@ -59,7 +59,7 @@ module.exports = (store) => {
         const compatibleVersion = (version === 6 && expected === 7) || (expected === 6 && version === 7)
 
         if (version !== expected && !compatibleVersion) {
-          return callback(new Error(`version mismatch: expected v${expected}, found v${version}`))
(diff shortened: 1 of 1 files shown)


src/version.js
```diff
@@ -59,7 +59,7 @@ module.exports = (store) => {
         const compatibleVersion = (version === 6 && expected === 7) || (expected === 6 && version === 7)
 
         if (version !== expected && !compatibleVersion) {
-          return callback(new Error(`version mismatch: expected v${expected}, found v${version}`))
+          return callback(new Error(`repository version mismatch: expected v${expected}, found v${version}`))
```
jacobheun: The js-ipfs issue is about clarity for the user. This is an improvement, but perhaps we should borrow the go language, and leave the expected and found version?

`ipfs repo needs migration: expected v${expected}, found v${version}`

While we don't yet support this in the browser it gives the user an action they can take to remedy the problem.

PR #185, opened 2018-12-10
fix: fix repo lock and buffer api

hugomrdias:
This PR makes the repo lock safer with callback/promise conversion, using then second arg we don't catch error inside the first arg fn and avoid callback already called errors.

Plus changes 2 two test to use the new buffer api.

### DIFF
--- a/src/lock.js
+++ b/src/lock.js
@@ -39,5 +39,8 @@ exports.lock = (dir, callback) => {
         }
       })
     })
-    .catch(err => callback(err))
+    .catch(err => {
+      log(err)
+      callback(err)
+    })
 }
--- a/src/lock.js
+++ b/src/lock.js
@@ -38,9 +38,8 @@ exports.lock = (dir, callback) => {
             .catch(err => cb(err))
         }
       })
-    })
+    }, callback)
     .catch(err => {
       log(err)
-      callback(err)
     })
 }


achingbrain:
What happens when the invocation of `callback` inside the `.then` throws?  I think you'll end up with an unhandled promise rejection:

```javascript
// index.js
const callback = () => {
  throw new Error('Derp')
}

const work = () => {
  return Promise.resolve()
}

work()
  .then(() => {
    callback()
  }, () => {
  console.info('never invoked')
})
```

```console
(node:72480) UnhandledPromiseRejectionWarning: Error: Derp
    at callback (/path/to/index.js:2:9)
    at work.then (/path/to/index.js:11:5)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:745:11)
    at startup (internal/bootstrap/node.js:266:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:596:3)
(node:72480) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:72480) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

hugomrdias:
yes true but at least get to see the error, right now the stack start at callOnlyOnce inside async and the actual error isn't there.
This will get better with the async iterators work and for now either we handle the error top level or if it throws like in your example its probably a bug on our side and we can find the real problem easier.

src/lock.js
```diff
@@ -36,6 +36,5 @@ exports.lock = (dir, callback) => {
           .then(() => cb())
           .catch(err => cb(err))
       }})
-    })
-    .catch(err => callback(err))
+    }, callback)
```
jacobheun: @hugomrdias let's add the catch and log the error for now to avoid the unhandled rejection.  We can improve this when we move async. This also needs to be rebased with master.

hugomrdias:
should be good now @jacobheun

jacobheun:
@hugomrdias your original intent was to avoid the callback twice error, correct? This is just logging whatever error we get now, but still has a risk of that. Shouldn't this look like the following? This will avoid the double callback and catch and log any error if it's thrown from the callback.

```js
  lock(dir, { lockfilePath: file, stale: STALE_TIME })
    .then(release => {
      callback(null, {
        close: (cb) => {
          release()
            .then(() => cb())
            .catch(err => cb(err))
        }
      })
    }, callback)
    .catch(err => {
      log(err)
    })
```

hugomrdias:
done

PR #186, opened 2018-12-22
fix: reduce bundle size

hugomrdias:
this PR doesn't actually reduces repo bundle size but help a lot the top level repo like js-ipfs


src/config.js
```diff
@@ -38,7 +37,7 @@ module.exports = (store) => {
         } catch (err) {
           return callback(err)
         }
-        if (key !== undefined && !_has(config, key)) {
+        if (key !== undefined && !_get(config, key)) {
```
jacobheun: If the `key` sent to `dlv` happens to not be a string, the original object might be returned. I think this is causing failures in the js-ipfs tests. We should probably convert / do some type checking.

src/config.js
```diff
@@ -37,11 +37,16 @@ module.exports = (store) => {
...
+          return callback(null, config)
+        }
+
+        if (typeof key !== 'string') {
+          return callback(new Error('Key ' + key + ' must be a string.'))
         }
-        const value = key !== undefined ? _get(config, key) : config
-        callback(null, value)
+
+        callback(null, _get(config, key, null) || new Error('Key ' + key + ' does not exist in config'))
```
jacobheun: The error needs to be passed in the first param of callback.

hugomrdias: OMG ofc

PR #189, opened 2019-02-11
Refactor js-repo away from callbacks leaning on async/await

zcstarr:
Initial pull request for the async await refactor. Some notes. 

I don't anticipate test to past just yet. There are upstream dependencies on async-multihashing and the various datastore interfaces. I modeled the interface changes based off of the current inflight async api refactor PRs for the dependencies(https://github.com/multiformats/js-multihashing-async/pull/37, https://github.com/ipfs/interface-datastore/pull/25, https://github.com/ipfs/js-datastore-level/pull/12). 

Another thing to note is that some of the error handling for opening/deleting a repo, might be up for question. It was a bit unclear what errors should be handled, given the ongoing refactor. 
https://github.com/ipfs/js-ipfs-repo/blob/41e042b29485b58f1c8191e8a4306bb40cc8a5a1/src/index.js#L71

Let me know what you think and what needs to be patched up or changed, happy to jam on it. If the interface seems good. I'll try to push testing the interface as far as I can go prior to the upstream package changes landing.


src/index.js
```diff
@@ -161,199 +127,149 @@ class IpfsRepo {
...
   /**
    * Check if the repo is already initialized.
-   *
    * @private
-   * @param {function(Error)} callback
-   * @returns {void}
+   * @returns {Promise<void>}
    */
-  _isInitialized (callback) {
+  async _isInitialized () {
```
alanshaw: should now return true/false

src/index.js
```diff
@@ -61,83 +56,61 @@ class IpfsRepo {
...
-    ], callback)
+    await this.root.open()
+    await this.config.set(buildConfig(config))
+    await this.spec.set(buildDatastoreSpec(config))
+    await this.version.set(repoVersion)
   }
 
   /**
    * Open the repo. If the repo is already open no action will be taken.
    * If the repo is not initialized it will return an error.
```
AuHau: I believe that "If the repo is already open no action will be taken." is not really correct, because according to the L77, it will throw an error. Maybe lets update it?

zcstarr: Good catch, the comment is definitely off 👍

AuHau: Maybe the error thrown should be bit more custom (eq. creating a new category in `./errors/index.js`) so people could detect if the opening of repo failed because it is locked and handle it? But not sure if it is in the scope of this PR though...

src/version.js
```diff
@@ -11,58 +11,44 @@ module.exports = (store) => {
...
      */
-    get (callback) {
-      store.get(versionKey, (err, buf) => {
-        if (err) {
-          return callback(err)
-        }
-        callback(null, parseInt(buf.toString().trim(), 10))
-      })
+    async get () {
+      const buf = await this.get(versionKey)
```
dirkmc: This line causes a stack overflow, I think it should be
```suggestion
      const buf = await store.get(versionKey)
```

zcstarr: It does, I have some pending changes stored locally for this, that I haven't pushed back up yet. Do you mind if I drop a PTAL when I push the various fixes I have going locally, this PR  was a bit tricker to write and is probably more error prone in places, because I wrote it  before the other PRs. Thanks for looking into it!

dirkmc: No worries, I just noticed this when testing some code that has this as a dependency. The rest of it looks great to me.

src/index.js
```diff
@@ -161,199 +134,137 @@ class IpfsRepo {
...
-        log('unlocking')
-        this.closed = true
-        this._closeLock(cb)
-      },
-      (cb) => {
-        this.lockfile = null
-        cb()
-      }
-    ], (err) => callback(err))
+    await this.apiAddr.delete()
```
AuHau: This refactored part will fail if the /api does not exists; see original `this.apiAddr.delete(ignoringNotFound(cb))`.

src/index.js
```diff
@@ -61,83 +56,61 @@ class IpfsRepo {
...
+  async init (config) {
     log('initializing at: %s', this.path)
-
-    series([
-      (cb) => this.root.open(ignoringAlreadyOpened(cb)),
-      (cb) => this.config.set(buildConfig(config), cb),
-      (cb) => this.spec.set(buildDatastoreSpec(config), cb),
-      (cb) => this.version.set(repoVersion, cb)
-    ], callback)
+    await this.root.open()
```
dirkmc:
```suggestion
    await this.root.open(ignoringAlreadyOpened())
```

src/index.js
```diff
@@ -61,83 +56,61 @@ class IpfsRepo {
...
+      this.datastore = backends.create('datastore', path.join(this.path, 'datastore'), this.options)
+      log('creating blocks')
+      const blocksBaseStore = backends.create('blocks', path.join(this.path, 'blocks'), this.options)
+      this.blocks = await blockstore(blocksBaseStore, this.options.storageBackendOptions.blocks)
+      log('creating keystore')
+      this.keys = backends.create('keys', path.join(this.path, 'keys'), this.options)
+      this.closed = false
+      log('all opened')
+    } catch (err) {
       if (err && this.lockfile) {
```
dirkmc: I think you still want to throw here if there is no lock file.

src/index.js
```diff
@@ -61,83 +56,61 @@ class IpfsRepo {
...
-          callback(err)
-        })
-      } else {
-        callback(err)
+        try {
+          this._closeLock()
+        } catch (err2) {
+          log('error removing lock', err2)
+        }
+        this.lockfile = null
```
dirkmc: I think you only want to set `this.lockfile` to null if there was not an error thrown from `this._closeLock()`

src/config.js
```diff
@@ -12,97 +11,75 @@ const configKey = new Key('config')
...
+      if (!key || typeof key !== 'string') {
+        throw new Error('Invalid key type')
       }
 
       if (value === undefined || Buffer.isBuffer(value)) {
-        return callback(new Error('Invalid value type'))
+        throw new Error('Invalid value type')
       }
 
       setQueue.push({
```
dirkmc: I'd suggest using a p-queue so that you can return a Promise here

package.json
```diff
@@ -49,15 +49,14 @@
     "rimraf": "^2.6.2"
   },
   "dependencies": {
-    "async": "^2.6.1",
     "base32.js": "~0.1.0",
     "bignumber.js": "^8.0.2",
     "cids": "~0.5.7",
     "datastore-core": "~0.6.0",
     "datastore-fs": "~0.7.0",
```
AuHau: Could you please add a dependency on https://github.com/ipfs/js-datastore-fs/pull/22 ?

package.json
```diff
@@ -48,28 +48,29 @@
...
-    "datastore-fs": "~0.8.0",
-    "datastore-level": "~0.10.0",
+    "datastore-core": "git://github.com/zcstarr/js-datastore-core.git",
+    "datastore-fs": "git://github.com/zcstarr/js-datastore-fs.git",
+    "datastore-level": "git://github.com/ipfs/js-datastore-level.git#refactor/async-iterators",
     "debug": "^4.1.0",
-    "dlv": "^1.1.2",
-    "interface-datastore": "~0.6.0",
+    "err-code": "^1.1.2",
+    "interface-datastore": "git://github.com/ipfs/interface-datastore.git#refactor/async-iterators",
```
AuHau: As the `refactor/async-iterators` PR was merged and the branch deleted this is outdated and breaks things.

achingbrain:
Superseded by #199

PR #191, opened 2019-02-14
chore: update deps

jacobheun:
The most notable update here is the move to `datastore-fs@0.8` which contains performance improvements.


jacobheun:
My plan is to do an `@next` release of this and do some testing with ipfs and interop before releasing the final version; per the note about testing, https://github.com/ipfs/js-datastore-fs/pull/21#issue-240442280.

PR #195, opened 2019-03-29
fix: memlock handling existing locks

AuHau:
Branch is based on the #189 PR, I will revisit this and rebase it once it is merged.
Also, the tests are not passing because of the current state of the base PR, I will revisit them too once the PR is merged.


AuHau:
Not actual anymore, new PR #200

PR #198, opened 2019-05-12
chore: update cids dependency

alanshaw:
Note that since this module does not return CIDs, this change could be released as a patch version.


jacobheun:
Released in v0.26.6

PR #199, opened 2019-05-30
Refactor/async await

achingbrain:
Supersedes #189 and includes all of the wonderful work that @zcstarr did in that PR.

All async/await refactors that this relies on have been merged & released.  I think this also now addresses all of the PR comments from #189 - can it be merged & released?


README.md
```diff
@@ -157,63 +147,61 @@ Arguments:
...
-#### `repo.close (callback)`
+#### `Promise repo.close ()`
 
 Unlocks the repo.
 
-#### `repo.exists (callback)`
+#### `Promise<boolean> repo.exists ()`
 
-Tells whether this repo exists or not. Calls back with `(err, bool)`.
+Tells whether this repo exists or not. Returns a boolean.
```
alanshaw:
```suggestion
Tells whether this repo exists or not. Returned promise resolves to a `boolean`.
```

README.md
```diff
@@ -157,63 +147,61 @@ Arguments:
...
 
-#### `repo.put (key, value:Buffer, callback)`
+#### `Promise repo.put (key, value:Buffer)`
 
 Put a value at the root of the repo.
 
 * `key` can be a buffer, a string or a [Key](https://github.com/ipfs/interface-datastore#keys).
 
-#### `repo.get (key, callback)`
+#### `Promise<value> repo.get (key)`
```
alanshaw: It's always a buffer right?
```suggestion
#### `Promise<Buffer> repo.get (key)`
```

README.md
```diff
@@ -222,77 +210,70 @@ This is contains a full implementation of [the `interface-datastore` API](https:
...
+assert.equal(config.a.b.c, 'c value')
 ```
 
-##### `repo.config.get(value, callback)`
+##### `Promise repo.config.set(value)`
 
 Set the whole config value. `value` can be any object that is serializable to JSON.
 
-##### `repo.config.get(key:string, callback)`
+##### `Promise<string> repo.config.get(key:string)`
```
alanshaw:
```suggestion
##### `Promise<?> repo.config.get(key:string)`
```

README.md
```diff
@@ -222,77 +210,70 @@ This is contains a full implementation of [the `interface-datastore` API](https:
...
 
-Get a config value. `callback` is a function with the signature: `function (err, value)`, wehre the `
-value` is of the same type that was set before.
+Get a config value. Returns the same type that was set before.
```
alanshaw:
```suggestion
Get a config value. Returned promise resolves to the same type that was set before.
```

README.md
```diff
@@ -222,77 +210,70 @@ This is contains a full implementation of [the `interface-datastore` API](https:
...
 
-### `repo.stat ([options], callback)`
+### `Promise<Object> repo.stat ([options])`
 
 Gets the repo status.
 
 `options` is an object which might contain the key `human`, which is a boolean indicating whether or not the `repoSize` should be displayed in MiB or not.
 
-`callback` is a function with the signature `function (err, stats)`, where `stats` is an Object with the following keys:
+Returns an Object with the following keys:
```
alanshaw:
```suggestion
Returned promise resolves to an `Object` with the following keys:
```

README.md
```diff
@@ -311,27 +292,27 @@ const memoryLock = require('ipfs-repo/src/lock-memory')  // Default in browser
...
 
-#### `lock.locked (dir, callback)`
+#### `Promise<boolean> lock.locked (dir)`
 
 Checks the existence of the lock.
 
 `dir` is a string to the directory to check for the lock. The repo typically checks for the lock at its root.
 
-`callback` is a function with the signature `function (err, boolean)`, where `boolean` indicates the existence of the lock.
+Returns a `boolean` indicating the existence of the lock.
```
alanshaw:
```suggestion
Returned promise resolves to a `boolean` indicating the existence of the lock.
```

package.json
```diff
@@ -42,34 +43,40 @@
...
     "just-safe-set": "^2.1.0",
+    "lodash.get": "^4.4.2",
+    "lodash.has": "^4.5.2",
+    "lodash.set": "^4.3.2",
     "multiaddr": "^6.0.6",
+    "p-queue": "^5.0.0",
     "proper-lockfile": "^4.0.0",
     "pull-stream": "^3.6.9",
-    "sort-keys": "^2.0.0"
+    "safe-buffer": "^5.1.2",
```
alanshaw: `buffer` _and_ `safe-buffer`?

package.json
```diff
@@ -42,34 +43,40 @@
...
     "debug": "^4.1.0",
     "dlv": "^1.1.2",
-    "interface-datastore": "~0.6.0",
+    "err-code": "^1.1.2",
+    "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
     "just-safe-set": "^2.1.0",
+    "lodash.get": "^4.4.2",
+    "lodash.has": "^4.5.2",
+    "lodash.set": "^4.3.2",
```
alanshaw: Use `just-safe-set` and `just-safe-get` for smaller bundle size?

src/blockstore.js
```diff
@@ -55,142 +47,117 @@ function createBaseStore (store) {
...
-     * @returns {void}
+     * @returns {Promise<void>}
      */
-    delete (cid, callback) {
+    delete (cid) {
       if (!CID.isCID(cid)) {
-        return setImmediate(() => {
-          callback(new Error('Not a valid cid'))
-        })
+        throw new Error('Not a valid cid')
```
alanshaw: Needs to be `async` or users can't `delete(cid).catch(console.log)` this error.

There's quite a few places where this happens - we should decide if allowing library consumers to use promise syntax is important.

cc @ipfs/wg-js-core

src/lock-memory.js
```diff
@@ -11,41 +10,34 @@ const LOCKS = {}
 
 /**
  * Lock the repo in the given dir.
- *
+ * TODO
```
alanshaw: What is TODO?

src/lock-memory.js
```diff
@@ -11,41 +10,34 @@ const LOCKS = {}
...
  * @param {string} dir
- * @param {function(Error, lock)} callback
- * @returns {void}
+ * @returns {Promise<Object>}
  */
-exports.lock = (dir, callback) => {
+exports.lock = (dir) => {
```
dirkmc: per @alanshaw's comment above, this should be `async`

test/options-test.js
```diff
@@ -42,21 +42,23 @@ describe('custom options tests', () => {
...
-      () => repo._openLock(repo.path)
-    ).to.throw('Locks must have a close method')
-    done()
+    let error
+    try {
+      await repo._openLock(repo.path)
+    } catch (err) {
+      error = err
+    }
+    expect(error.message).to.equal('Locks must have a close method')
```
dirkmc: Should we test this with an error code instead?

src/config.js
```diff
@@ -1,121 +1,89 @@
...
-
-        const value = _get(config, key, null)
-
-        if (value === null) {
-          return callback(new Error('Key ' + key + ' does not exist in config.'))
-        }
+      const encodedValue = await store.get(configKey)
+      const config = JSON.parse(encodedValue.toString())
+      if (key !== undefined && !_has(config, key)) {
+        throw new Error(`Key ${key} does not exist in config`)
```
dirkmc: Should we add some tests for these error cases?

src/blockstore.js
achingbrain: Marking things async creates a Promise.  Best case scenario there's no actual async in the function, but the VM will still put the code after the promise onto the microtask queue which creates a small amount of latency.  On hot codepaths this adds up as the Nearform performance analysis of our codebase showed.  Our linting rules enforce [`require-await`](https://eslint.org/docs/rules/require-await) for this reason (and would cause the suggested change to fail linting).

I take the point about promise syntax - it's not great to have an API where some functions are async and other aren't, but we should not create artificial asynchronicity for the reason outlined above.

Maybe I'm wrong but I'd consider this module an internal API, quite different to `js-ipfs` itself so the inconsistency bothers me, but not to the point that we should make the code less efficient.

package.json
achingbrain: The bundled produced is 20kb smaller with `just-safe-set` and `just-safe-get` instead of `lodash.get` and `lodash.set`.

src/lock-memory.js
achingbrain: It's a mystery. Have removed it.

src/index.js
```diff
@@ -157,204 +125,173 @@ class IpfsRepo {
...
 
+  /**
+   * Opens the root backend, catching and ignoring an 'Already open' error
+   * @returns {Promise}
+   */
+  async _openRoot () {
+    try {
+      await this.root.open()
+    } catch (err) {
+      if (err.message !== 'Already open') {
```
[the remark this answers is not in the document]
achingbrain: That would be good.  TBH I'm not sure where this message originates from - the default root datastore implementation is `datastore-fs` which does not throw an error with the message 'Already Open'.

There are [these standardised errors](https://github.com/ipfs/interface-datastore/blob/master/src/errors.js), maybe the best thing to do is to open a PR that expands that list with more error types?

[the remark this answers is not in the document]
dirkmc: Thinking a little more about it, it seems like we often throw an error when reopening an already open thing, or reclosing an already closed thing, and then we have to write code like this with guards for that same error. It may make things simpler if we just ignore a double-open or double-close in our libs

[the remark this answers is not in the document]
achingbrain: Sounds sensible, though is outside the scope of an callback->async/await refactor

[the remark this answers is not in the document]
dirkmc: Fair enough 👍

package.json
achingbrain: Have removed both.

src/blockstore.js
alanshaw: I'm curious to know if `await syncFn()` has the same effect.

We're always going to call functions that may or may not be async with `await` right? - because we just don't know if the function will return a promise or not.

So, if `await` also forces async behaviour then is there a benefit to not using `async`?

alanshaw: > If the value of the expression following the await operator is not a Promise, it's converted to a resolved Promise.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#Description

src/config.js
achingbrain: I have increased the test coverage.  The only branches that are not covered now stem from [this line](https://github.com/ipfs/js-ipfs-repo/blob/d52ecc09bd0c69dbab439c559624f3d2d223edac/src/blockstore.js#L175) which I'm not sure how to test, given that the `cid` passed to it has already been run through `CID.isCID()` so nothing in that function should throw.

achingbrain: I have increased the test coverage.

src/blockstore.js
achingbrain: Exactly why you shouldn't just `await` on everything 😁

FWIW the future-tech version of IPLD figures out if the return value of a function is `then`able and resolves it if so (or at least it did in one iteration).  Making everything async by default prevents these sorts of optimisations.

Anyway I've updated the PR, everything in the docs that claims to be async is now marked async.

achingbrain:
@jacobheun any comments or can this be merged & released?

jacobheun:
I haven't had a chance to go through this in depth but I also don't need to be a blocker on this since @dirkmc and @alanshaw reviewed.

I can do a merge and minor release of this, the integration with js-ipfs should also help catch any items that may have been missed.

jacobheun:
@achingbrain v0.27.0 is out.

Issue #197, opened 2019-05-10
Config.get(...) should raise interface-datastore.errors.notFound 


AuHau:
I am not 100% sure about this, but while working on integrating the `js-ipfs-repo-migrations` I stumbled on a case where I have expected to `ipfs-repo` throw an `interface-datastore.errors.notFound` error when `config.get()` won't find the required key. 

I now that `interface-datastore.errors` are related to datastores and this is only somehow related to it as `ipfs-repo.config` creates an abstraction layer on top of it. But it would be good to have some easy and clean way how to test for *not found* errors other than `err.message.match('does not exist')`.

AuHau:
Solved by #201 just not by throwing `interface-datastore.errors.notFound` but by throwing custom error from repo level.

PR #201, opened 2019-06-27
Introducing not found error for config values

AuHau:


AuHau:
I have now realized that earlier I created issue #197 that propose the usage of `interface-datastore.errors.notFound` instead of this custom error.

@jacobheun what is your take on this?

jacobheun:
@AuHau they should probably be separate. While they both have the same error code at the moment, the missing config key is not a database error. There's no guarantee the database code won't change in the future, so keeping them separate feels like a safer way to avoid issues down the road.

The `err-code` module is already being used in other places in ipfs-repo, is there any particular reason you're using error classes instead? If not it would probably be good to use the former.

AuHau:
Sounds good! 

Well, it is a style of errors proposed in https://github.com/ipfs/js-ipfs/pull/1746 that I silently highly prefer 😅 and I hope is the future in js-ipfs land. Should I refactor it nevertheless?

jacobheun:
My only real hesitation with including them is that we have 2 ways errors are being created now. I know there is another PR open that adds another code. Would you be up for submitting another PR to replace the rest of the err-code usage?

AuHau:
Sure! I will happily do that. It should not be even breaking change as the `code` properties will stay the same. I will make it happen tomorrow.

PR #202, opened 2019-06-28
Automatic repo migrations

AuHau:
This PR brings integration of js-ipfs-repo-migrations brings automatic repo migrations to ipfs-repo (both in-browser and fs). It is possible to control the automatic migration using either config's setting 'repoDisableAutoMigration' or IPFSRepo's option 'disableAutoMigration'.

Related PRs: [js-ipfs](https://github.com/ipfs/js-ipfs/pull/2044), [js-ipfs-repo-migrations](https://github.com/AuHau/js-ipfs-repo-migrations/pull/1)

### Questions to tackle:
- [x] should throw err when version is higher then expected? (currently only log)
- [x] `repo.version.check()` change of behaviour (throw ==> bool)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
-    "ipfs-repo-migrations": "ipfs/js-ipfs-repo-migrations#dev",
+    "ipfs-repo-migrations": "ipfs/js-ipfs-repo-migrations#master",
     "just-safe-get": "^1.3.0",
     "just-safe-set": "^2.1.0",
     "lodash.has": "^4.5.2",
--- a/package.json
+++ b/package.json
@@ -43,16 +43,16 @@
     "npm": ">=3.0.0"
   },
   "devDependencies": {
-    "aegir": "^19.0.3",
+    "aegir": "^20.4.1",
     "chai": "^4.2.0",
     "dirty-chai": "^2.0.1",
     "lodash": "^4.17.11",
     "memdown": "^4.0.0",
-    "multihashes": "~0.4.14",
-    "multihashing-async": "~0.7.0",
+    "multihashes": "~0.4.15",
+    "multihashing-async": "~0.8.0",
     "ncp": "^2.0.0",
-    "rimraf": "^2.6.3",
-    "sinon": "^7.3.1"
+    "rimraf": "^3.0.0",
+    "sinon": "^7.5.0"
   },
   "dependencies": {
     "base32.js": "~0.1.0",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
-    "ipfs-repo-migrations": "github:ipfs/js-ipfs-repo-migrations#master",
+    "ipfs-repo-migrations": "~0.1.0",
     "just-safe-get": "^1.3.0",
     "just-safe-set": "^2.1.0",
     "lodash.has": "^4.5.2",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
-    "ipfs-repo-migrations": "AuHau/js-ipfs-repo-migrations#dev",
+    "ipfs-repo-migrations": "ipfs/js-ipfs-repo-migrations#dev",
     "just-safe-get": "^1.3.0",
     "just-safe-set": "^2.1.0",
     "lodash.has": "^4.5.2",
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
-    "ipfs-repo-migrations": "ipfs/js-ipfs-repo-migrations#master",
+    "ipfs-repo-migrations": "github:ipfs/js-ipfs-repo-migrations#master",
     "just-safe-get": "^1.3.0",
     "just-safe-set": "^2.1.0",
     "lodash.has": "^4.5.2",
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "multihashes": "~0.4.14",
     "multihashing-async": "~0.7.0",
     "ncp": "^2.0.0",
-    "rimraf": "^2.6.3"
+    "rimraf": "^2.6.3",
+    "sinon": "^7.3.1"
   },
   "dependencies": {
     "base32.js": "~0.1.0",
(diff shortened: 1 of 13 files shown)


src/index.js
```diff
@@ -264,6 +275,40 @@ class IpfsRepo {
...
+    const currentRepoVersion = await this.version.get()
+    log(currentRepoVersion)
+    if (currentRepoVersion >= toVersion) {
+      if (currentRepoVersion > toVersion) {
+        log('Your repo\'s version is higher then this version of js-ipfs-repo require! You should revert it.')
+      }
+
+      log('Nothing to migrate')
+      return
+    }
```
alanshaw: I'm not sure I understand this - I think this check needs to be done in the migrator.

If `currentRepoVersion` === `toVersion` then nothing to do, no error.
If `currentRepoVersion` > `toVersion` then it needs downgrade, error if no downgrade path.

package.json
```diff
@@ -73,6 +75,7 @@
   },
   "license": "MIT",
   "contributors": [
+    "Adam Uhlir<adam@uhlir.dev>",
```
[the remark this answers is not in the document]
AuHau: Oh didn't know that, thanks!

package.json
```diff
@@ -64,6 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
+    "ipfs-repo-migrations": "AuHau/js-ipfs-repo-migrations#dev",
```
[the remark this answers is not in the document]
AuHau: Yeah that is in plan (I should have add WIP to this PR as it waits for this last step. sorry).

src/default-options-browser.js
```diff
@@ -2,6 +2,7 @@
 
 // Default configuration for a repo in the browser
 module.exports = {
+  disableAutoMigration: false,
```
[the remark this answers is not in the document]
AuHau: Yeah I definitely like that better, I just followed what was said in the original issue, where this was proposed.

src/index.js
AuHau: Yeah, this definitely can be more streamlined. 

Sooo you are saying that it should automatically also downgrade? This is currently not happening and from my experience with migration tools, it is also usually not the case. There often is a way how to downgrade (and our migrator supports it), but it is not automatically integrated. I think that the main reason is to provide easy way "forward", but way "back" is usually something that most users won't need.

I checked `go` and they don't supports it.

package.json
```diff
@@ -64,6 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
+    "ipfs-repo-migrations": "ipfs/js-ipfs-repo-migrations#master",
```
daviddias: This can point to the released version now, correct?

AuHau: Yes, I still have not released it as I want to have LGTM for the PRs in case of some changes would have to be made. It will be the last step before merge.

src/index.js
AuHau: True about the browsers.

To your question, yes we can, but not always. Some migrations will be irreversible and in that case, an exception will be the answer I guess.

src/index.js
```diff
@@ -264,6 +275,33 @@ class IpfsRepo {
...
+      autoMigrateConfig = await this.config.get(AUTO_MIGRATE_CONFIG_KEY)
+    } catch (e) {
+      if (e.code === ERRORS.NotFoundError.code) {
+        autoMigrateConfig = true // Config's default value is True
+      } else {
+        throw e
+      }
+    }
+
+    return autoMigrateConfig && this.options.autoMigrate
```
alanshaw: I think this is not deciding if it should migrate correctly! Or at least not consistently with how we usually override config.

1. `true` in config and `true` in options = `true` ✅
1. `true` in config and `false` in options = `false` ✅
1. `true` in config and `undefined` in options = `undefined` ❌ (expected `true`)
1. `false` in config and `true` in options = `false` ❌ (expected `true`)
1. `false` in config and `false` in options = `false` ✅
1. `false` in config and `undefined` in options = `false` ✅
1. `undefined` in config and `true` in options = `undefined` ❌ (expected `true`)
1. `undefined` in config and `false` in options = `undefined` ✅
1. `undefined` in config and `undefined` in options = `undefined` ❌ (expected `true`)

For 3, 6, and 8: in the code `this.options.autoMigrate` will normally be `true`/`false` unless the user passes `{ autoMigrate: undefined }`.

The behaviour I was expecting was:

If the user explicitly passed an option to `autoMigrate` as `true`/`false` then that is the value that should be used. If not explicitly passed as an option then we should check for a value in repo config otherwise we use the default (true).

The problem comes when the user doesn't pass a value for `autoMigrate` - by the time we get to this function we're already defaulted it to `true`/`false` without checking for a value in the repo config. Might be best to not have a value set for this in `defaultOptions` and do a check like:

```suggestion
    return this.options.autoMigrate || autoMigrateConfig || true
```

...although you can probably skip fetching `autoMigrateConfig` from the repo if `this.options.autoMigrate` is `true` or `false`.

**Can you please ensure there are tests to assert this behaviour is working as expected?**

test/node.js
```diff
@@ -87,7 +112,10 @@ describe('IPFS Repo Tests onNode.js', () => {
     })
 
     after(async () => {
-      await repo.close()
+      try {
+        await repo.close()
+      } catch (e) {
```
alanshaw: Why does this need a try/catch now?

test/node.js
```diff
@@ -16,8 +17,32 @@ const fsstat = promisify(fs.stat)
...
+
+  const repo = new IPFSRepo(repoPath, opts)
+
+  if (init) {
+    await repo.init({})
+  } else {
+    await asyncNcp(testRepoPath, repoPath)
+  }
+
+  if (!dontOpen) {
```
alanshaw: As far as I can tell `init`, `dontOpen` and `opts` are not used?

test/node.js
```diff
@@ -16,8 +17,32 @@ const fsstat = promisify(fs.stat)
...
+    await repo.open()
+  }
+
+  return {
+    path: repoPath,
```
alanshaw: ...and neither this? How about just return the repo instance?

alanshaw:
ping @AuHau

AuHau:
Sorry, got a job now, so it might take some time to get back to this. Hopefully in the following days.

package.json
```diff
@@ -64,6 +65,7 @@
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
     "ipfs-block": "~0.8.1",
+    "ipfs-repo-migrations": "github:ipfs/js-ipfs-repo-migrations#master",
```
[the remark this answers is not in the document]
AuHau: Hmmmmmmmmmm, I think that with the automatic downgrading we will have to wait to release this with some proper migration (eq. the migration-8 which is in the process). Because now `js-ipfs-repo-migration` reports the latest migration it has as `0` because there are no migrations. Now in this auto-downgrade behavior, it will try to downgrade to `0` which will fail........ 🙈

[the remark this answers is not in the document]
alanshaw: Add an empty v7 migration?

test/node.js
AuHau: Yeah, I think that I wanted to refactor the tests somehow using this function, but did not finished it 😅I will just remove it...

AuHau:
I believe this is ready to be merged.
`js-ipfs-repo-migrations` is released and updated here.

alanshaw:
@jacobheun please can we have review+merge+release?

jacobheun:
@alanshaw on it!

jacobheun:
0.29 is out! 🚀

PR #203, opened 2019-08-07
refactor: error objects and handling

AuHau:
This refactors style of error objects that are thrown and that follows style proposed in https://github.com/ipfs/js-ipfs/pull/1746/

I would say big plus of this PR is that it moves all errors to `./errors` and hence create overviews of errors that are thrown from `ipfs-repo`.

@jacobheun I am not so sure about this not introducing breaking change any more. The thrown errors indeed contain `code` property as mentioned earlier, but the problem is how to assert the code's value. Earlier you could do with some errors: 
```js
const errors = require('ipfs-repo').errors

if(err.code === errors.ERR_REPO_ALREADY_OPEN)
...
```

now you have to do:
```js
const errors = require('ipfs-repo').errors

if(err.code === errors.ERR_REPO_ALREADY_OPEN.code)
...
```

Which packages uses `js-ipfs-repo`? I assume `js-ipfs`? I will create PR to reflects this change there, but should I look somewhere else?

### DIFF
--- a/src/errors/index.js
+++ b/src/errors/index.js
@@ -1,5 +1,7 @@
 'use strict'
 
+// TODO: Should be refactored when https://github.com/ipfs/js-ipfs/pull/1746/ is adopted.
+
 /**
  * Error raised when there is lock already in place when repo is being opened.
  */
@@ -13,7 +15,7 @@ class LockExistsError extends Error {
 }
 
 LockExistsError.code = 'ERR_LOCK_EXISTS'
-exports.LockExistsError = LockExistsError
+exports.ERR_LOCK_EXISTS = LockExistsError
 
 /**
  * Error raised when requested item is not found.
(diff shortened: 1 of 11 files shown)


src/errors/index.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 
+// TODO: Should be refactored when https://github.com/ipfs/js-ipfs/pull/1746/ is adopted.
```
AuHau: In the PR there are some utils to ease the creation of the Error's class, but since I believe it will be moved out to some general package to be used across `js` land, I did not copy&pasted it directly here and just created the Error's classes manually. Once it will land this can be adopted here.

jacobheun:
@hugomrdias can you weigh in here since you're looking at changes in js-ipfs errors. The error exports are going to be a breaking change. If we're breaking the api, I'd like to avoid doing so more than once with these updates.

AuHau:
I also had an idea in order to not introduce breaking change, to export
these new error objects under some other property (for example:
future_errors or errs) keep the previous one and only mark them as
deprecated and replace them with the new objects only after the mass
adoption of this error style across JS land. What do you think?

Adam Uhlíř

(Sent from mobile phone)

jacobheun:
Closing this until error code handling is determined in js-ipfs https://github.com/ipfs/js-ipfs/pull/1746

PR #208, opened 2019-09-09
feat: trying to callbackify ipfs-repo

vmx:
This callbackified ipfs-repo. Almost all tests pass, except the ones for
custom locking.

Though when used with js-ipfs, the js-ipfs tests won't work. It seems
that starting up a node is stuck somewhere.


jimpick:
I’ll take a look at the js-ipfs tests.

achingbrain:
Since https://github.com/ipfs/js-ipfs/pull/2390 was merged, is this still necessary?

achingbrain:
This is no longer necessary as we are full async/await everywhere now.

PR #209, opened 2019-09-20
feat: accept async iterators into blockstore.putMany

achingbrain:
Allows streaming into the blockstore. The blockstore currently uses the batch functionality of the underlying datatore, so we might need to change `interface-datastore` next to add streaming primitives but this at least allows streaming this far down the stack.

BREAKING CHANGE: you must pass an iterable or async iterable to putMany - this should be relatively painless as the current API is to pass an array which is iterable, but it does change the API and rules is rules.

### DIFF
--- a/src/blockstore.js
+++ b/src/blockstore.js
@@ -85,26 +85,21 @@ function createBaseStore (store) {
     /**
      * Like put, but for more.
      *
-     * @param {Array<Block>} blocks
+     * @param {AsyncIterable<Block>} blocks
      * @returns {Promise<void>}
      */
     async putMany (blocks) {
-      const keys = blocks.map((b) => ({
-        key: cidToKey(b.cid),
-        block: b
-      }))
-
       const batch = store.batch()
 
-      await Promise.all(
-        keys.map(async k => {
-          if (await store.has(k.key)) {
-            return
-          }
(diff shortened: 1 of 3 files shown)


src/blockstore.js
```diff
@@ -85,26 +85,21 @@ function createBaseStore (store) {
     /**
      * Like put, but for more.
      *
-     * @param {Array<Block>} blocks
+     * @param {AsyncIterable<Block>} blocks
```
hugomrdias:
```suggestion
     * @param {AsyncIterable<Block>|Iterable<Block>} blocks
```

src/blockstore.js
```diff
@@ -85,26 +85,21 @@ function createBaseStore (store) {
     /**
      * Like put, but for more.
      *
-     * @param {Array<Block>} blocks
+     * @param {AsyncIterable<Block>|Iterable<Block>} blocks
      * @returns {Promise<void>}
      */
     async putMany (blocks) {
```
jacobheun: The Readme needs to be updated to reflect the change, otherwise this looks good.

achingbrain: Done!

jacobheun:
Released in v0.28.0

PR #211, opened 2019-10-03
feat: store blocks under multihash key

AuHau:
This is related to https://github.com/ipfs/js-ipfs/issues/2415

### Breaking change

`repo.blocks.query()` now returns multihashes as a key instead of CID.

### Question!

`repo.blocks.query()` currently returns Datastore's query object (eq. `{key: '', value: ''}`). Shouldn't it return `js-ipfs-block`'s Block instead? All other `repo.blocks` functions return Block... As I will be refactoring all usages of `repo.blocks.query()` this might be good oportunity to change it.

Yet if we would do it, we would have to reconstruct the CID using IPLD RAW codec. This will be generally desired behavior of the `query()` function anyway (there is already implemented behavior for this, but is used only when called as `query({}, true)`).

But on the other hand not sure how we would handle the case when `query()` would be called with for example `repo.blocks.query({keysOnly: true})`...


daviddias:
> repo.blocks.query() currently returns Datastore's query object (eq. {key: '', value: ''}). Shouldn't it return js-ipfs-block's Block instead? All other repo.blocks functions return Block... As I will be refactoring all usages of repo.blocks.query() this might be good oportunity to change it.

Sounds about right! :) Good catch!

AuHau:
@daviddias alright, but then  what about the `repo.blocks.query({keysOnly: true})` use-case? From what I saw in the code actually most of the usage of `repo.blocks.query()` only needs keys and not the values.
If we would use `Block()` then it would not be possible.

PR #213, opened 2019-11-07
fix: human readable option

PedroMiguelSS:
This PR fixes the output of `human` option that can be passed to `repo.stat ([options])`. 

Both `repoSize` and `storageMax` properties of the returned object are now according to `go` implementation:
- `repoSize` -> represented in `MB`
- `storageMax` -> represented in `GB`

Example:
```js
const stats = await repo.stat({ human: true })
console.log(stats)
// Logs:
//{
//  repoPath: '/var/folders/l1/l5dg09hj7x9d6yxm5z2vsr2h0000gn/T/test-repo-for-1573141229356',
//  storageMax: '10 GB',
//  version: 7,
//  numObjects: 24,
//  repoSize: '0.07 MB'
//}
```

Closes https://github.com/ipfs/js-ipfs/issues/1996.


src/index.js
```diff
@@ -308,7 +313,7 @@ class IpfsRepo {
   async _storageMaxStat () {
     try {
       const max = await this.config.get('Datastore.StorageMax')
-      return new Big(max)
+      return new Big(bytes(max))
```
achingbrain: Why is this necessary?

src/index.js
```diff
@@ -7,6 +7,8 @@ const debug = require('debug')
 const Big = require('bignumber.js')
 const errcode = require('err-code')
 const migrator = require('ipfs-repo-migrations')
+const filesize = require('filesize')
```
achingbrain: `pretty-bytes` is more than half the size of `filesize`.  Any reason not to use it?

src/index.js
```diff
@@ -258,19 +260,22 @@ class IpfsRepo {
...
+        ? filesize(storageMax, { exponent: 3 }) // exponent 3 specifies the symbol GB for base 2
+        : storageMax,
       version: version,
-      numObjects: blocks.count,
-      repoSize: size
+      numObjects: options.human
+        ? blocks.count.toNumber()
+        : blocks.count,
+      repoSize: options.human
+        ? filesize(size, { exponent: 2 }) // exponent 2 specifies the symbol MB for base 2
```
achingbrain: My opinion on this is not terribly strong but I guess it'd be nice to let the library figure out a suitable representation - if the repo size is 10KB, show 10KB instead of 0.01MB

src/index.js
PedroMiguelSS: No reason at all. We're using `pretty-bytes` now.

src/index.js
PedroMiguelSS: This is necessary because `datastore` config can have a default value as you can see [here](https://github.com/ipfs/js-ipfs-repo/blob/master/test/test-repo/config#L9) and this need to be parsed to bytes. Otherwise, `Big` object won't be properly instantiated.

PedroMiguelSS:
Can I get a second review here, please? @achingbrain @hugomrdias @jacobheun

jacobheun:
Released in v0.29.1

alanshaw:
I'd like to keep formatting functionality out of ipfs core - it bloats the bundle size, makes assumptions about locale and can be easily added by applications using IPFS.

What do you think? I'd be in favour of removing this functionality altogether and pushing it into the CLI and HTTP API resources.

PedroMiguelSS:
I got your point. It makes sense. What do you think @achingbrain @jacobheun?

jacobheun:
> What do you think? I'd be in favour of removing this functionality altogether and pushing it into the CLI and HTTP API resources.

This sounds like a good approach to me.

achingbrain:
Me too, formatting is a UI concern.

PR #214, opened 2019-11-14
fix: close root datastore

hugomrdias:
blocks, keys and datastore are being closed in repo.close() but root is not.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -228,7 +228,7 @@ class IpfsRepo {
       }
     }
 
-    await Promise.all([this.blocks, this.keys, this.datastore].map((store) => store.close()))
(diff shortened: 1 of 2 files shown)


hugomrdias:
tests are weird everything works on the same repo, i can't just stop instances

any tips where to write the test ?

hugomrdias:
This came from ctl refactor, i was trying to clean the disposable browser nodes and kept getting an error. Then i found that cleaning in the browser is completely different from node, one needs to delete 4 idb's and i could delete 3 of them but not the `root` one because it wasn't being close here.

jacobheun:
@hugomrdias can you fix the linting issue with CI?

PR #215, opened 2019-11-25
fix: pass backwards-compatible level-js options

achingbrain:
Between `level-js@2` and `level-js@4` it started passing version numbers into  `indexedDB.open` so we have to specify these now.  It also prefixes locations with a disambiguator, which we need to set to blank for browsers to still be able to read their previously created databases.

Depends on:

- [x] https://github.com/Level/packager/pull/96
- [x] https://github.com/Level/level-js/pull/184
- ~[ ] https://github.com/ipfs/js-ipfs-repo/pull/217~

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "cids": "~0.7.0",
     "datastore-core": "~0.7.0",
     "datastore-fs": "~0.9.0",
-    "datastore-level": "~0.12.0",
+    "datastore-level": "~0.14.0",
     "debug": "^4.1.0",
     "err-code": "^1.1.2",
     "interface-datastore": "~0.7.0",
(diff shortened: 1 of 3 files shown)


vweevers:
> it started passing version numbers

Hm, this was actually intended to maintain backwards compatibility. Because before `level-js@3`, it used [`IDBWrapper`](https://github.com/jensarps/IDBWrapper) which also passed a version number into `indexedDB.open`.

> It also prefixes locations with a disambiguator, which we need to set to blank for browsers to still be able to read their previously created databases.

In addition, `IDBWrapper` had a default prefix, so databases created with `level-js < 3` do in fact have a prefix.

achingbrain:
@jacobheun if you're happy with this could you please release it & backport it to 0.29.x so we can do a patch release for ipfs@0.39.0 and fix ipfs/js-ipfs#2629?

PR #216, opened 2019-11-25
feat: remove options object from stat method

PedroMiguelSS:
`stat()` method was accepting an object called `options`. The only option available was `human` that was changing the stats' output to be human-readable. 

`human` option was just a formatting functionality and it should be kept out of `ipfs-core`. See https://github.com/ipfs/js-ipfs-repo/pull/213#issuecomment-555488344 for further details.

Thus, this PR removes the `options` object.


jacobheun:
v0.30.0 is on the webs. 🕸

PR #217, opened 2019-11-29
chore: update deps

achingbrain:


achingbrain:
Hmm, actually hold off this for the minute, it seems we might need to more in `datastore-level` to read databases created with earlier versions of `level-js`.

achingbrain:
I've pulled the `datastore-level` update out of this PR as it's handled in #215, this should be good to go.

PR #219, opened 2020-02-09
feat: add isInitialized method

achingbrain:
A few places need ot know whether or not a repo has been initialised and do so by a variety of methods - checking whether certain files exist, etc.

This PR adds an `isInitialized()` method to the ipfs repo class to take some of this guesswork away as the repo will know if it's been initialised.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -74,13 +74,21 @@ class IpfsRepo {
    * @returns {Promise<Boolean>}
    */
   async isInitialized () {
+    if (!this.closed) {
+      // repo is open, must be initialized
+      return true
+    }
+
     try {
+      // have to open the root datastore in the browser before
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -69,6 +69,22 @@ class IpfsRepo {
...
 
+  /**
+   * Check if the repo is already initialized.
+   * @returns {Promise<Boolean>}
+   */
+  async isInitialized () {
+    try {
+      await this._openRoot()
+      await this._checkInitialized()
+      // necessary? await this.root.close()
```
alanshaw: Is there a way to detect initialization without opening the root store?

achingbrain: Only in node, in the browser you have to open the store first.

PR #221, opened 2020-02-12
feat: store pins in datastore

achingbrain:
Adds a `pins` datastore to store pins in.


test/pins-test.js
```diff
@@ -0,0 +1,79 @@
...
+  describe('pins', () => {
+    const dataList = range(100).map((i) => Buffer.from(`hello-${i}-${Math.random()}`))
+    const data = Buffer.from('hello world')
+    const b = new Key('hello')
+
+    it('exists', () => {
+      expect(repo).to.have.property('pins')
+    })
+
+    it('implements interface-datastore', () => {
```
vasco-santos, line 20: This test is redundant as `datastore-level` already runs the interface tests: https://github.com/ipfs/js-datastore-level/blob/master/test/index.spec.js#L44

PR #225, opened 2020-04-02
feat: use new datastore-idb

hugomrdias:
This PR removes node globals and reduces bundle size. size: 77.09KB -> 45.07KB
Also a new fast and smaller browser datastore is used `datastore-idb`  https://github.com/ipfs/js-datastore-idb
```
datastore-idb up to 44% faster and 69% smaller

batch idb x 5.26 ops/sec ±4.57% (29 runs sampled)
batch level x 2.92 ops/sec ±3.76% (19 runs sampled)
Fastest is batch idb
    
size 37.32KB to 11.41KB
```


achingbrain:
Can this be split into two PRs please?  One for node globals (a patch) and one for a new datastore (a major)

hugomrdias:
can't remove node globals with datastore-level

package.json
```diff
@@ -43,33 +43,35 @@
...
+    "sinon": "^9.0.2"
   },
   "dependencies": {
-    "base32.js": "~0.1.0",
     "bignumber.js": "^9.0.0",
+    "buffer": "^5.5.0",
     "bytes": "^3.1.0",
-    "cids": "^0.8.0",
-    "datastore-core": "~0.7.0",
+    "cids": "~0.8.0",
```
achingbrain, line 61: Why has this changed from `^` to `~`?

test/lock-test.js
```diff
@@ -17,25 +18,12 @@ module.exports = (repo) => {
...
       } catch (err) {
-        if (process.listeners('uncaughtException').length > 0) {
-          expect(err.message).to.match(/already locked|already held|already being held|ELOCKED/)
-        }
-      } finally {
-        // Reset listeners to maintain test integrity
-        process.removeAllListeners('uncaughtException')
-        process.addListener('uncaughtException', mochaExceptionHandler)
+        expect(err.code)
+          .to.be.eq(LockExistsError.code)
```
achingbrain, line 26:
```suggestion
          .to.equal(LockExistsError.code)
```

package.json
hugomrdias: i don't know probably happened in the rebase, but ~ is what we use for below  1.0.0 right ?

package.json
```diff
@@ -43,33 +43,35 @@
...
     "datastore-level": "~0.14.0",
     "debug": "^4.1.0",
     "err-code": "^2.0.0",
-    "interface-datastore": "^0.8.0",
-    "ipfs-block": "~0.8.1",
-    "ipfs-repo-migrations": "~0.1.0",
+    "interface-datastore": "^0.8.3",
+    "ipfs-repo-migrations": "^0.2.0",
+    "ipfs-utils": "^2.2.0",
+    "ipld-block": "^0.9.1",
```
achingbrain, line 71: This is probably going to need a follow-up PR to js-ipfs at least that swaps ipfs-block for ipld-block

hugomrdias: yes! the following PR in js-ipfs to remove node globals will include that

package.json
achingbrain: We used to have a linting rule that enforced that, then we relaxed it because npm uses `^` for above 0.0.x

hugomrdias: i have changed it back

hugomrdias:
> LGTM, just needs green CI

yeah trying to fix it... i have no idea whats happening with the linux runs

test/interop-test.js
```diff
@@ -27,7 +27,7 @@ module.exports = (repo) => {
 
       const values = await Promise.all(cids.map((cid) => repo.blocks.get(cid)))
       expect(values.length).to.equal(2)
-      expect(values.map((value) => value.data.length)).to.eql([2659, 12783])
+      expect(values.map((value) => value.data.length)).to.eql([2665, 12789])
```
achingbrain, line 30: Why did these change?

PR #229, opened 2020-04-30
feat: adds .replace and .getAll methods to config

achingbrain:
The config .get/.set methods are overloaded to allow the user to
retrieve or replace the entire config.  This is an error prone
approach that could end with the user invalidating their config when
they did not mean to.

The change here adds `.replace` and `.getAll` methods to allow the
user to be specific about their intentions and hopefully makes the
API slightly less of a foot-gun.


achingbrain:
This is a backport to the `1.x.x` release.

PR #232, opened 2020-05-07
feat: use streaming API for blockstore access

achingbrain:
Uses the streaming APIs from interface-datastore to work with data being sent/fetched from the blockstore.

The previous behaviour was to use batches for putting streams of blocks - this was bad because we'd end up buffering the whole stream before adding blocks to the blockstore.  Instead now we stream the blocks as we get them.

* Adds `repo.blockstore.getMany()` function
* Passes `options` to underlying datastores to allow passing AbortSignals

BREAKING CHANGES:

* `repo.blockstore.putMany()` used to return a promise that resolved when all the deletes were done, now it returns an AsyncIterator that yields `{key, value}` objects as the put progresses
* `repo.blockstore.deleteMany()` used to return a promise that resolved when all the deletes were done, now it returns an AsyncIterator that yields CID objects as the delete progresses


src/blockstore.js
```diff
@@ -52,88 +52,107 @@ function createBaseStore (store) {
...
       }
 
       const k = cidToKey(block.cid)
-      const exists = await store.has(k)
-      if (exists) return
-      return store.put(k, block.data)
+      const exists = await store.has(k, options)
+
+      if (exists) {
+        return
```
hugomrdias, line 91: this should probably return block also

achingbrain: Done!

PR #297, opened 2021-04-07
feat: make blockstore identity-hash compatible

hannahhoward:
# Goals

Make the blockstore properly return blocks for "identity" CIDs (CID's where the multihash algorithm is "identity", meaning the contents of the hash are the actual data for the block, and there is no other data associated with the CID).

# Implementation

- Separate out typing for the blockstore interface
- Implement `idstore` based on https://github.com/ipfs/go-ipfs-blockstore/blob/master/idstore.go - this leaves the original blockstore implementation unchanged but adds a composable layer applied on top to add compatibility with identity hashes (served as the same blockstore interface)
- add `idstore` in construction of `repo.blocks`
- fix minor compatibility issues with `just-safe-get` and `just-safe-set` by fixing to patch release for now

# For discussion

I based implementation for delete and put operations on the go versions, which essentially treats a put or delete for an identity hash as a no-op. However, the golang version does not return the block itself for puts, and I wasn't sure whether I should be doing that. For now, I've made it so the identity block is included in the results of puts.

Since I extracted `Blockstore` typings to an interface, to save writing them twice for `blockstore` and `idstore` and insure they both return the exact same types, I removed the JSDoc comments on `blockstore`. I wasn't sure though if that affects what shows up in JSDocs. (I did put the type on the return value for createBaseStore)

fix https://github.com/ipfs/js-ipfs/issues/3289


package.json
```diff
@@ -81,12 +81,14 @@
     "ipfs-repo-migrations": "^7.0.1",
     "ipfs-utils": "^6.0.0",
     "ipld-block": "^0.11.0",
+    "it-filter": "^1.0.2",
     "it-map": "^1.0.2",
     "it-pushable": "^1.4.0",
-    "just-safe-get": "^2.0.0",
-    "just-safe-set": "^2.1.0",
+    "just-safe-get": "~2.0.0",
```
achingbrain, line 87: What was the error that meant this needed to change from `^` to `~`?

src/idstore.js
```diff
@@ -0,0 +1,140 @@
...
+ * @typedef {import("interface-datastore").Options} DatastoreOptions
+ * @typedef {import('./types').Blockstore} Blockstore
+ */
+
+/**
+ *
+ * @param {Blockstore} blockstore
+ */
+module.exports = (blockstore) => {
+  return createIdStore(blockstore)
```
achingbrain, line 23: Why not just export the `createIdStore` function?

achingbrain (changes requested):

achingbrain:
Nice work, thanks for submitting this.

To land this could you please open a PR against js-ipfs that depends on this PR as a gh dep and add tests for identity CIDs to the interface-ipfs-core suite?  At least for [dag.get](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/dag/get.js) and [block.get](https://github.com/ipfs/js-ipfs/blob/master/packages/interface-ipfs-core/src/block/get.js) but anywhere these code paths will be executed. 
 This will make sure it's tested against core, over http and that it behaves in the same way as go-IPFS (over http).

There's an overview of how to run the tests here: https://github.com/ipfs/js-ipfs/blob/master/docs/DEVELOPMENT.md#run-tests

hannahhoward:
@achingbrain I've responded to PR comments and reset the dependencies, resolving the breaking change in the latest `just-safe-get` / `just-safe-set` by fixing our calls to match the newly added types. A PR with tests to js-ipfs is incoming.

hannahhoward:
PR for js-ipfs to demonstrate fix: https://github.com/ipfs/js-ipfs/pull/3616

achingbrain (approved):

PR #302, opened 2021-04-14
chore: update types

achingbrain:
- Updates deps
- Changes Blockstore.query return type based on query arg
- Exports Blockstore type from index.js
- Moves Config types from ipfs-core-types to ipfs-repo
- Removes bignumber.js dep in favour of using `BigInt`s
- Normalises type import style
- Removes ts-ignores from modules with types


src/blockstore.js
```diff
@@ -45,6 +45,7 @@ function createBaseStore (store) {
       return store.open()
     },
 
+    // @ts-ignore TODO: how to not have ts freak out about this
```
vasco-santos, line 48: What is the issue here?

src/idstore.js
```diff
@@ -30,6 +30,8 @@ function createIdStore (store) {
     open () {
       return store.open()
     },
+
+    // @ts-ignore TODO: how to not have ts freak out about this
```
vasco-santos, line 34: ..

src/config.js
```diff
@@ -43,7 +44,7 @@ module.exports = (store) => {
      * @param {string} [key] - the config key to get
      * @param {Object} [options] - options
      * @param {AbortSignal} [options.signal] - abort this config read
-     * @returns {Promise<unknown>}
+     * @returns {Promise<Config | any>}
```
vasco-santos, line 47: Can't we get `any` out of here?

src/types.d.ts
```diff
@@ -110,4 +106,99 @@ export interface Blockstore {
...
+  Gateway?: string
+  Swarm: string[]
+}
+
+export interface APIConfig {
+  HTTPHeaders?: Record<string, string>
+}
+
+export interface DiscoveryConfig {
+  MDNS: MDNSDiscovery
```
vasco-santos, line 141: is the discovery config required per module?

src/types.d.ts
```diff
@@ -110,4 +106,99 @@ export interface Blockstore {
...
-}
\ No newline at end of file
+}
+
+export interface Config {
+  Addresses: AddressConfig
+  API?: APIConfig,
+  Profiles?: string
+  Bootstrap: string[]
+  Discovery: DiscoveryConfig
```
vasco-santos, line 116: is the discovery config required ?

src/types.d.ts
```diff
@@ -110,4 +106,99 @@ export interface Blockstore {
...
+export interface WebRTCStarDiscovery {
+  Enabled?: boolean
+}
+
+export interface DatastoreConfig {
+  Spec?: DatastoreSpec
+}
+
+export interface DatastoreSpec {
+
```
vasco-santos, line 159: Missing Datastore spec?

src/blockstore.js
achingbrain: It can't work out that we are yielding only CIDs or Blocks and not a mixture of both so we are, in fact, compliant with the interface.

Maybe it could use a better comment 😆

src/config.js
achingbrain: If you omit the `key` you get `Config`, otherwise you get `any` because it could be a string or an array or and object, etc.  We could make `key` mandatory but they we'd just lose `Config` as a return type.  Might be worth doing anyway TBF.

achingbrain: I've made `key` mandatory, since we don't treat it as optional in js-ipfs.

src/types.d.ts
achingbrain: The options are different between MDNS and WebRTCStar so it seems so, yes

src/types.d.ts
achingbrain: No, looks like a mistake

src/types.d.ts
achingbrain: I've updated it

src/blockstore.js
achingbrain: I've added a better comment

src/config.js
```diff
@@ -34,52 +34,50 @@ module.exports = (store) => {
      * @param {AbortSignal} [options.signal] - abort this config read
      * @returns {Promise<Config>}
      */
-    getAll (options = {}) { // eslint-disable-line require-await
-      return configStore.get(undefined, options)
+    async getAll (options = {}) { // eslint-disable-line require-await
+      // level-js@5.x cannot read keys from level-js@4.x dbs so fall back to
```
vasco-santos, line 38: Is there an issue to track this? If not, can you create one?

vasco-santos (approved):
LGTM

src/config.js
achingbrain: I can do, but it's not something that's ever going to be fixed so the issue will hang around forever.

The 4 -> 5 version bump was because of this breaking change.

vasco-santos: ok makes sense, so maybe not worth

PR #308, opened 2021-04-22
chore: fix types

achingbrain:


vasco-santos:
the types check is reporting a problem with the `keysOnly` property

vasco-santos (approved):

PR #313, opened 2021-05-14
chore: upgrade repo to new multiformats module

achingbrain:
Replaces `cids`, `multibase`, `multihashes` etc with new `multiformats` module.

Depends on:

- [x] https://github.com/ipfs/interface-datastore/pull/91
- [x] https://github.com/ipfs/js-ipfs-repo-migrations/pull/98

BREAKING CHANGE: The blockstore now takes instances of the new `CID` class and returns `Uint8Arrays`


vasco-santos (changes requested):

src/blockstore-utils.js
```diff
@@ -13,11 +14,13 @@ const uint8ArrayToString = require('uint8arrays/to-string')
  * @returns {Key}
  */
 exports.cidToKey = cid => {
-  if (!CID.isCID(cid)) {
+  if (!(cid instanceof CID)) {
```
rvagg, line 17: I think the appropriate pattern to be using with the new `CID` is this:

```js
cid = CID.asCID(cid)
if (cid == null) { throw ... }
```

I don't think standard will let you `if ((cid = CID.asCID(cid)) != null)` but that'd be a neat shortcut.

It does an `instanceof` with a fast pass-through as the first check, but if that doesn't work it'll try a few other things and get us past the usual `instanceof` and version problems, looking for a _CID-like_, then return `null` if it can't make it work.

src/blockstore.js
```diff
@@ -39,28 +38,56 @@ function maybeWithSharding (filestore, options) {
...
     },
 
+    close () {
+      return store.close()
+    },
+
     async * query (query, options) {
-      for await (const { key, value } of store.query(query, options)) {
-        yield new Block(value, keyToCid(key))
+      /** @type {import('interface-datastore').Query} */
```
vasco-santos, line 52:
```suggestion
      /** @type {import('interface-blockstore').Query} */
```

Shouldn't be the blockstore query?

src/blockstore.js
```diff
@@ -1,16 +1,15 @@
...
 
 const { shard, ShardingDatastore } = require('datastore-core')
-const Block = require('ipld-block')
 const { cidToKey, keyToCid } = require('./blockstore-utils')
 const drain = require('it-drain')
 const pushable = require('it-pushable')
+
 /**
- * @typedef {import('interface-datastore').Query} Query
+ * @typedef {import('multiformats').CID} CID
```
rvagg, line 9: A very minor nit - when importing just one thing, you can reach into that particular export, in this case `import('multiformats/cid').CID` gets you the same thing but makes a much narrower import path. Probably doesn't matter a whole lot here for type imports (it may be a tiny bit faster?) but it ends up being very efficient for ESM-aware bundling or smarter tree-shaking bundlers so it's nice to demonstrate it to anyone looking at the code.

src/blockstore.js
```diff
@@ -39,28 +38,56 @@ function maybeWithSharding (filestore, options) {
...
+      }
+
+      for await (const { key, value } of store.query(storeQuery, options)) {
+        yield { key: keyToCid(key), value }
       }
     },
 
     async * queryKeys (query, options) {
-      for await (const key of store.queryKeys(query, options)) {
+      /** @type {import('interface-datastore').KeyQuery} */
```
vasco-santos, line 71:
```suggestion
      /** @type {import('interface-blockstore').KeyQuery} */
```

src/idstore.js
```diff
@@ -1,18 +1,17 @@
 'use strict'
 
-const Block = require('ipld-block')
 const filter = require('it-filter')
-const mh = require('multihashes')
 const pushable = require('it-pushable')
 const drain = require('it-drain')
-const CID = require('cids')
+const { CID } = require('multiformats')
```
rvagg, line 6: you should be able to `require('multiformats/cid')` here for the same reason mentioned above, avoid the root export if/where possible if you only want one thing

src/blockstore.js
```diff
@@ -151,8 +172,14 @@ function createBaseStore (store) {
       return out
     },
 
-    close () {
-      return store.close()
+    batch () {
```
vasco-santos, line 175: Don't we support batch? If not, perhaps throw an error instead?

src/idstore.js
```diff
@@ -1,18 +1,17 @@
 'use strict'
 
-const Block = require('ipld-block')
```
rvagg, line 3: there is a `Block` in multiformats, it's not quite as featureful as ipld-block IIRC or what we tried with https://github.com/ipld/js-block, but it's a really nice container `cid:bytes:value` that can be quite useful with some extra sugar that might be close to what the user previously expected. We could return that to the user and they get to consume it as they want.

See https://github.com/multiformats/js-multiformats/blob/master/src/block.js (I don't think it got documented yet because we were working on getting it _right_ and figuring out the tradeoffs between that and what we might do in js-block if we did that above it, but it should be good to use as is).

src/idstore.js
```diff
@@ -123,21 +133,18 @@ function createIdStore (store) {
  * @returns {{ isIdentity: false } | { isIdentity: true, digest: Uint8Array}}
  */
 function extractContents (k) {
-  if (!CID.isCID(k)) {
+  if (!(k instanceof CID)) {
```
rvagg, line 136: `asCID`

test/blockstore-test.js
```diff
@@ -38,15 +44,16 @@ module.exports = (repo) => {
...
+    let pair
     /** @type {CID} */
     let identityCID
+
     before(async () => {
-      const hash = await multihashing(bData, 'sha2-256')
-      b = new Block(bData, new CID(hash))
-      const identityHash = await multihashing(identityData, 'identity')
-      identityCID = new CID(1, 'dag-cbor', identityHash)
+      const digest1 = await sha256.digest(bData)
```
rvagg, line 53: You could use `require('multiformats/block').Block` here, e.g.:

```js
const block = await Block.encode({ value: bData, codec: identity, hasher: sha256 })
// then you have:
// block.cid
// block.bytes
// block.value
```

test/blockstore-utils-test.js
```diff
@@ -3,14 +3,15 @@
 
 const { expect } = require('aegir/utils/chai')
 const { Key } = require('interface-datastore')
-const CID = require('cids')
+const { CID } = require('multiformats')
```
rvagg, line 6: `'multiformats/cid'`

test/interop-test.js
```diff
@@ -2,34 +2,34 @@
 'use strict'
 
 const { expect } = require('aegir/utils/chai')
-const mh = require('multihashing-async').multihash
-const CID = require('cids')
+const { CID } = require('multiformats')
```
rvagg, line 5: `'multiformats/cid'`

test/interop-test.js
```diff
@@ -2,34 +2,34 @@
...
 /**
  * @param {import('../src/index')} repo
  */
 module.exports = (repo) => {
   describe('interop', () => {
     it('reads welcome-to-ipfs', async () => {
-      const welcomeHash = mh.fromHexString(
-        '1220120f6af601d46e10b2d2e11ed71c55d25f3042c22501e41d1246e7a1e9d3d8ec'
+      const welcomeHash = CID.decode(
+        uint8ArrayFromString('1220120f6af601d46e10b2d2e11ed71c55d25f3042c22501e41d1246e7a1e9d3d8ec', 'base16')
```
rvagg, line 17: there might be potential to trim dependencies a bit more, multiformats exports a `bytes` that has some basic utilities including `fromHex()` and `toHex()` (just available on the root export though):

```js
const { CID, bytes } = require('multiformats')
// ...

CID.decode(bytes.fromHex('1220120f...'))
```

test/interop-test.js
```diff
@@ -2,34 +2,34 @@
...
       )
 
-      const val = await repo.blocks.get(new CID(welcomeHash))
-      expect(uint8ArrayToString(val.data)).to.match(/Hello and Welcome to IPFS/)
+      const val = await repo.blocks.get(welcomeHash)
+      expect(uint8ArrayToString(val)).to.match(/Hello and Welcome to IPFS/)
```
rvagg, line 21: `bytes` also has a `fromString()` and `toString()` that does utf-8 with `TextEncoder` and `TextDecoder` similar to this btw

test/stat-test.js
```diff
@@ -2,20 +2,18 @@
 'use strict'
 
 const { expect } = require('aegir/utils/chai')
-const Block = require('ipld-block')
-const CID = require('cids')
+const { CID } = require('multiformats')
```
rvagg, line 5: `'multiformats/cid'`

rvagg:
Cool, nice bundle min reduction too

As mentioned in the inline comments, it might be worth exploring `require('multiformats/block').Block` as a replacement for ipld-block. It's lightweight and can be just used for `cid`, `bytes` and `value` properties but has some extras that might make it a decent replacement.

/cc @gozala who might like to see the code being consumed here and also have thoughts on usage of `Block`

src/blockstore-utils.js
rvagg:
```suggestion
  cid = CID.asCID(cid)
  if (cid == null) {
```

src/blockstore.js
rvagg:
```suggestion
 * @typedef {import('multiformats/cid').CID} CID
```

src/idstore.js
rvagg:
```suggestion
const { CID } = require('multiformats/cid')
```

src/idstore.js
rvagg:
```suggestion
  cid = CID.asCID(cid)
  if (cid == null) {
```

test/blockstore-utils-test.js
rvagg:
```suggestion
const { CID } = require('multiformats/cid')
```

test/interop-test.js
rvagg:
```suggestion
const { CID } = require('multiformats/cid')
```

test/stat-test.js
rvagg:
```suggestion
const { CID } = require('multiformats/cid')
```

rvagg:
followed @vasco-santos's lead with suggestions for easy merge if that helps (not sure it does) for the easy ones

src/blockstore.js
achingbrain: I merged this, but it's wrong - the blockstore wraps a datastore so you have to convert the blockstore query into a datastore query (e.g. the key types are different)

achingbrain:
>  it might be worth exploring `require('multiformats/block').Block`

I've got my eye on it but so far it hasn't been necessary - CIDs in, Uint8Arrays out seems to cover all of our use cases.

Gozala:
> /cc @Gozala who might like to see the code being consumed here and also have thoughts on usage of `Block`

I always felt that custom Blockstore API that was slightly incompatible with Datastore interface was awkward & I am glad to see it go. I think been at layer below `Block` here is a right decision.

